PR #121, opened 2024-03-20
feat: create Chart comppnent and extract all login from main page

### LINKED ISSUE #110: Extract all the chart logic to a component
This component should do the resizing and take props which are the series (graph values). When the values change, it will redraw the chart. So it will look like this:

```
const labels = ['January', 'February', 'March', 'April', 'May', 'June', 'July']
const series1 =  {
      label: 'My First dataset',
      fill: true,
      lineTension: 0.3,
      backgroundColor: 'rgba(225, 204,230, .3)',
      borderColor: 'rgb(205, 130, 158)',
      borderCapStyle: 'butt',
      borderDash: [],
      borderDashOffset: 0.0,
      borderJoinStyle: 'miter',
      pointBorderColor: 'rgb(205, 130,1 58)',
      pointBackgroundColor: 'rgb(255, 255, 255)',
      pointBorderWidth: 10,
      pointHoverRadius: 5,
      pointHoverBackgroundColor: 'rgb(0, 0, 0)',
      pointHoverBorderColor: 'rgba(220, 220, 220,1)',
      pointHoverBorderWidth: 2,
      pointRadius: 1,
      pointHitRadius: 10,
      data: [65, 59, 80, 81, 56, 55, 40],
}

const series2 = {
    {
      label: 'My Second dataset',
      fill: true,
      lineTension: 0.3,
      data: [28, 48, 40, 19, 86, 27, 90],
    },

const series3 = {
    {
      label: 'My Third dataset',
      data: [19, 86, 27, 90, 50, 80, 100],
    },

<Chart
    labels={labels}
    series={[series1, series2, series3]}
/>
```

Tip. Look at `$effect` rune and how to do `afterUpdate` with it https://svelte-5-preview.vercel.app/docs/runes#$effect and how to do `onDestroy`

### LINKED ISSUE #111: Remove unnecessary state variables that can live inside $effect
The `$effect` return function gets run:

1. immediately before the effect re-runs
2. when the component is destroyed

This means that if we have for example some debounce mechanism (e.g. the one added in #109), then we don't need to clear it in the effect function body only in the effect return function. We can also define the interval directly in the effect rather than have it as state.

See one the last examples in `$effect` docs https://svelte-5-preview.vercel.app/docs/runes#$effect

hlavkat:
resolve #110

### DIFF
--- a/src/lib/components/chart.svelte
+++ b/src/lib/components/chart.svelte
@@ -1,0 +1,56 @@
+<script lang="ts">
+	import Chart, { type ChartDataset } from 'chart.js/auto'
+
+	interface Props {
+		labels: number[]
+		series: ChartDataset[]
+	}
+	let { labels, series }: Props = $props()
+
+	let canvas: HTMLCanvasElement | null = $state(null)
+	let chart: Chart | null = $state(null)
+
+	$effect(() => {
+		if (canvas && !chart) {
+			chart = new Chart(canvas, {
+				type: 'line',
+				data: {
+					labels: labels,
+					datasets: series,
+				},
+
+				options: {
+					scales: {
+						y: {
+							min: 0,
+						},
+					},
+					responsive: false,
+				},
+			})
+			chart.resize()
+		}
+		if (chart) {
+			chart.data.labels = labels
+			chart.data.datasets = series
+			chart.update()
+		}
+	})
+	let prevChartWidth: number = $state(0)
+	let actChartWidth: number = $state(0)
+	$effect(() => {
+		const interval = setInterval(() => {
+			if (actChartWidth !== prevChartWidth && chart) {
+				prevChartWidth = actChartWidth
+				chart.resize()
+			}
+		}, 1000)
+		return () => {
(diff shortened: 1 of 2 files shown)


src/lib/components/chart.svelte
```diff
@@ -0,0 +1,56 @@
+<script lang="ts">
+	import Chart, { type ChartDataset } from 'chart.js/auto'
+
+	interface Props {
+		labels: number[]
```
vojtechsimetka:
```suggestion
		labels: (number | string)[]
```

src/lib/components/chart.svelte
```diff
@@ -0,0 +1,56 @@
...
+			}
+		}, 1000)
+		return () => {
+			clearInterval(interval)
+		}
+	})
+</script>
+
+<div class="chart">
+	<canvas bind:this={canvas} bind:clientWidth={actChartWidth} />
```
vojtechsimetka: this should be on the chart div

PR #122, opened 2024-03-21
feat: show warning on planned deposits/withdrawals

### LINKED ISSUE #113: Show warning on deposits/withdrawals that happened before date of birth or after end of investment horizon
This should look like helper text on the user inputs. The text should explain correctly what is wrong and should start with a the `Warning` icon

hlavkat:
resolve #113

### DIFF
--- a/src/lib/components/operation.svelte
+++ b/src/lib/components/operation.svelte
@@ -4,6 +4,7 @@
 	import { detailStore } from '$lib/stores/details.svelte'
 	import { year } from '$lib/utils'
 	import { Warning } from 'carbon-icons-svelte'
+	import { incrementDate } from '$lib/calc'
 
 	interface Props {
 		operation: Deposit | Withdrawal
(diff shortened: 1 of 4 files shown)


src/lib/components/operation.svelte
```diff
@@ -23,22 +26,49 @@
 		}
 		return ''
 	}
+	let operationErrors: string[] = $state([])
+	$effect(() => {
+		const dateOfBirth = new Date(detailStore.dateOfBirth).getTime()
+		const dateOfEndInvestment = dateOfBirth + detailStore.endAge * year
```
vojtechsimetka: Please use function `incrementDate` (it is in utils). The solution you suggested does not count with leap years etc.

src/lib/components/operation.svelte
```diff
@@ -23,22 +26,49 @@
...
+		if (dateOfStartOperation < dateOfBirth && dateOfEndInvestment < dateOfEndOperation) {
+			operationErrors = [$_('operationBeforeBorn'), $_('operationAfterEnd')]
+		} else if (dateOfStartOperation < dateOfBirth) {
+			operationErrors = [$_('operationBeforeBorn')]
+		} else if (dateOfEndInvestment < dateOfEndOperation) {
+			operationErrors = [$_('operationAfterEnd')]
+		} else {
+			operationErrors = []
+		}
+	})
```
vojtechsimetka: please change this to derived store instead of effect and state

src/lib/locales/cs.json
```diff
@@ -48,6 +48,8 @@
 	"plannedWithdrawals": "Plánované výběry",
 	"paidOnFees": "Zaplaceno na poplatcích",
 	"remainingPortfolioValue": "Zbývající částka portfolia",
+	"operationBeforeBorn": "Operace před narozením",
+	"operationAfterEnd": "Operace po planovaném konci",
```
vojtechsimetka: we should think about clearer error message

vojtechsimetka (changes requested):

PR #128, opened 2024-03-21
fix: replace store graph value as individual series instead of block

### LINKED ISSUE #125: Store graph values as individual series instead of block
Instead of 
```ts
interface GraphRecord {
	date: Date
	totalInvested: number
	totalDeposited: number
	totalWithdrawn: number
	totalFees: number
	totalDepositFee: number
	totalWithdrawFee: number
	totalManagementFee: number
	totalSuccessFee: number
}

...

let graphData: GraphRecord[] = $state([])
```

We want to store the individual series as arrays. So
```ts
dates: Date[] = $state([])
totalInvested: number[]
...
```

With this change we can remove the value mapping here https://github.com/vojtechsimetka/planuj.finance/blob/main/src/routes/%2Bpage.svelte#L121-L128

and  us the arrays directly in the `Chart` component https://github.com/vojtechsimetka/planuj.finance/blob/main/src/routes/%2Bpage.svelte#L248-L283

hlavkat:
resolve #125

### DIFF
--- a/src/lib/components/chart.svelte
+++ b/src/lib/components/chart.svelte
@@ -2,7 +2,7 @@
 	import Chart, { type ChartDataset } from 'chart.js/auto'
 
 	interface Props {
-		labels: number[]
+		labels: (number | string)[]
 		series: ChartDataset[]
 	}
 	let { labels, series }: Props = $props()
--- a/src/lib/components/chart.svelte
+++ b/src/lib/components/chart.svelte
@@ -2,7 +2,7 @@
 	import Chart, { type ChartDataset } from 'chart.js/auto'
 
 	interface Props {
-		labels: (number | string)[]
+		labels: number[]
 		series: ChartDataset[]
 	}
 	let { labels, series }: Props = $props()
--- a/src/lib/stores/results.svelte.ts
+++ b/src/lib/stores/results.svelte.ts
@@ -8,19 +8,16 @@ import {
 import { formatDate } from '$lib/utils'
 import { detailStore } from './details.svelte'
 
-interface GraphRecord {
-	date: Date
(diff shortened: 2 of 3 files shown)


src/lib/components/chart.svelte
```diff
@@ -2,7 +2,7 @@
 	import Chart, { type ChartDataset } from 'chart.js/auto'
 
 	interface Props {
-		labels: (number | string)[]
+		labels: number[]
```
vojtechsimetka: Lets keep it as it was before :)
```suggestion
		labels: (number | string)[]
```

src/lib/stores/results.svelte.ts
```diff
@@ -100,18 +103,17 @@ export function withResultsStore() {
...
-				totalInvested: newTotalInvested.toNumber(),
-				totalDeposited: newTotalDeposited.toNumber(),
-				totalWithdrawn: newTotalWithdrawn.toNumber(),
-				totalFees: newTotalFees.toNumber(),
-				totalManagementFee: newTotalManagementFee.toNumber(),
-				totalSuccessFee: newTotalSuccessFee.toNumber(),
-				totalDepositFee: newTotalDepositFee.toNumber(),
-				totalWithdrawFee: newTotalWithdrawFee.toNumber(),
-			})
+			if (i.getDate() % 30 === 0) {
```
vojtechsimetka: This is probable better - the values will be store for every first day of the month.
```suggestion
			if (i.getDate()  === 0) {
```

vojtechsimetka (changes requested):
Great progress, thanks a lot. I found few small issues I'd like you to handle.

PR #208, opened 2024-08-16
feat: update versions

agazso:
This PR updates the dependencies to use the latest version that Diete uses, including Diete components and css file. It does not change the code for now, so at the moment everything should work, just with the updated components and look.


vojtechsimetka (approved):
Nice it all works, it just made the app a bit more difficult to navigate as the placeholders are no longer labels so it is unclear what the inputs are. Also the readonly input fields at the end of the main page are now selectable which they were not before. I am fine if we fix this in a separate PR.

Screenshot 2024-08-19 at 8 21 11

Screenshot 2024-08-19 at 8 21 23

PR #214, opened 2024-08-20
feat: client creation flow

agazso:
This PR adds the client creation flow on the UI. The backend part is not yet implemented, that will be in a separate PR.

Notes:
- The paddings are not 100% following the design. The reason is that Diete does not specify paddings that the components follow, but rather each component can have custom paddings. It would be good to fix this in Diete.
- There is no header yet. It makes sense to add it when there is a database and a flow for signing up/logging in.
- The upload image is not working yet, because there is no backend atm.
- There is an issue with the date picker in a modal dialog


agazso:
Date picker in modal:

gnome-shell-screenshot-87gvor

vojtechsimetka (approved):

src/lib/locales/cs.json
```diff
@@ -1,4 +1,20 @@
...
+	"clientName": "TODO",
+	"datePlaceholder": "TODO",
+	"profilePicture": "TODO",
+	"addClient": "TODO",
+	"noClientsYet": "TODO",
+	"createYourFirstClient": "TODO",
+	"profileImageHelper": "TODO",
+	"uploadImage": "TODO",
+	"createClient": "TODO",
+
```
vojtechsimetka: You don't need to define these, it will use the EN by default and there is a tool that gives missing translations so it is easy to identify

PR #215, opened 2024-08-23
feat: supabase database 

hlavkat:
login and addClient is resolved

- [x] image of client must be resolved
- [x] the row level security must be resolved
- [x] migration


.vscode/extensions.json
```diff
@@ -0,0 +1,3 @@
+{
```
vojtechsimetka: Please remove this file

.vscode/settings.json
```diff
@@ -1,4 +1,10 @@
 {
```
vojtechsimetka, line 1: Please remove the changes to this file

vojtechsimetka (commented):
Please document how to run supabase, and how to create migrations from local DB, reset DB and other common tasks we do. Use the `pnpm supabase`

.gitignore
```diff
@@ -8,3 +8,4 @@ node_modules
 !.env.example
 vite.config.js.timestamp-*
 vite.config.ts.timestamp-*
+/supabase/temp
```
vojtechsimetka: We can remove this because folder `supabase` already has `.gitignore` containing these things.

README.md
```diff
@@ -36,3 +36,105 @@ npm run build
...
+
+### 1.Initialize Supabase in your project:
+
+-If you haven’t already initialized Supabase, navigate to your project directory and run:
+
+```bash
+pnpm supabase init
+```
+
+-This command creates a .supabase directory with configuration files.
```
vojtechsimetka: This step should be removed, we already have initialized supabase (there is  `supabase` directory).

vojtechsimetka (commented):
This should be fine now. Please review @agazso

src/lib/adapters/index.ts
```diff
@@ -0,0 +1,17 @@
+import type { ClientNoId } from '$lib/types'
+import SupabaseAdapter from './supabase'
+
+export interface Adapter {
+	// This is run when the app is mounted and should start app wide subscriptions
+	start?: () => Promise<void> | void
+	// This is run when the app unmounts and should clear subscriptions
+	stop?: () => Promise<void> | void
```
agazso: I would just make these explicit and if the actual adapter has an empty implementation it's fine.

```suggestion
	start: () => Promise<void>
	// This is run when the app unmounts and should clear subscriptions
	stop: () => Promise<void>
```

src/lib/components/input/input.svelte
```diff
@@ -13,7 +13,7 @@
 		dimension?: Dimension
 		layout?: Layout
 		unit?: string
-		error?: Snippet
+		error?: string[]
```
agazso: If there are more than one error then I would call this errors, and if there are no errors then it would be an empty list.

```suggestion
		errors: string[]
```

src/lib/components/input/input.svelte
```diff
@@ -102,7 +102,7 @@
 		</div>
 		{#if error}
 			<div class="error-message">
-				{@render error()}
+				{error}
```
agazso: This will just print strings unformatted, like `a,b,c`.
```suggestion
				{error.join(', ')}
```

src/lib/adapters/supabase/index.ts
```diff
@@ -0,0 +1,162 @@
...
+				} else {
+					console.log('User status changed', event)
+				}
+			}, 0)
+		})
+
+		this.authSubscription = subscription.data.subscription
+
+		onDestroy(() => {
+			this.authSubscription?.unsubscribe()
```
agazso: Why is this here and not in `stop()`?

vojtechsimetka: Yeah should be there and `stop()` should not be called elsewhere in the adapter

src/lib/components/input/input.svelte
vojtechsimetka: Actually, I don't like that this differs from diete.design. Maybe we should create snippets where we use this.

src/lib/types.ts
```diff
@@ -24,3 +26,17 @@ export const supportedCurrenciesWithLabels: CurrencyWithLabel = {
...
+export interface Store<T> {
+	data: T[]
+	loading: boolean
+}
+
+export interface Advisor {
+	email: string
+	password: string
+	confirmPassword: string
+}
```
agazso: If this is only used for form validation then I would make a zod schema out of this and call it `advisorFormSchema`, just like the others.

If it is not only used for form validation, then I would not have the `confirmPassword` here.

src/lib/components/input/input.svelte
agazso: Or maybe we should fix this in diete too

src/lib/adapters/supabase/index.ts
```diff
@@ -0,0 +1,160 @@
...
+	stop() {
+		this.authSubscription?.unsubscribe()
+		this.authSubscription = undefined
+
+		this.clientsSubscription?.unsubscribe()
+		this.clientsSubscription = undefined
+	}
+
+	async signUp(email: string, password: string) {
+		const { error } = await supabase.auth.signUp({ email, password })
```
agazso, line 119: Not sure how exactly this works, but the frontend should not have the credentials to be able to create a user. Instead it should call a backend that have the credentials to do this.

src/lib/components/input/input.svelte
vojtechsimetka: I could imagine diete using multiple options there (`Snippet`, `string` or `string[]`) but I would definitely keep `Snippet` which allows error styling.

src/lib/adapters/supabase/index.ts
vojtechsimetka: This is the same as firebase. You can create user from frontend no problem. But you can't view, edit or delete other users. Those are admin functions. But self registration is not. BTW there can be an approval mechanism, but I don't think we would need that.

src/lib/components/login.svelte
```diff
@@ -0,0 +1,94 @@
...
+	})
+</script>
+
+<div class="login">
+	<div class="login-form">
+		<Input
+			bind:value={email}
+			label="Email"
+			type="email"
+			error={formErrors?.email?._errors ?? undefined}
```
agazso: This only makes sense if `_errors` is `null` for some reason. If that's not case it's best to omit it.
```suggestion
			error={formErrors?.email?._errors}
```

src/lib/components/login.svelte
```diff
@@ -0,0 +1,94 @@
...
+		></Input>
+		<Input
+			bind:value={password}
+			type="password"
+			label="Password"
+			error={formErrors?.password?._errors ?? undefined}
```
agazso: This only makes sense if `_errors` is `null` for some reason. If that's not case it's best to omit it.
```suggestion
			error={formErrors?.password?._errors}
```

src/lib/schemas.ts
```diff
@@ -69,3 +69,24 @@ export const detailStoreSchema = portfolioSchema.extend({
...
 export const investmentSchema = z.object({})
+
+export const loginFormSchema = z.object({
+	email: z.string().email({ message: 'Email musí být platný.' }),
+	password: z.string().min(6, { message: 'Heslo musí mít alespoň 6 znaků.' }),
+})
+
+export const registerFormSchema = z
+	.object({
+		email: z.string().email({ message: 'Email musí být platný.' }),
```
agazso: Please don't write czech language error messages. The best is to use an identifier and add that to `en.json` or `cs.json` in `src/lib/locales`

agazso:
BTW docker desktop prints this in a log:
```
1970-01-01 01:00:00 invalid project name "kalkul.app": must consist only of lowercase alphanumeric characters, hyphens, and underscores as well as start with a letter or number
```
Everything works fine otherwise.

agazso (approved):

PR #224, opened 2024-09-06
feat: create portfolio flow

agazso:
This PR adds the create portfolio flow. No investments, they are going to be done in a separate PR.

Notes:
- Added migration and english translations.
- Moved the diete components under `$lib/components/ui`
- Removed some unused code that seemed more effort to fix.


supabase/migrations/20240906163118_create_portfolio.sql
```diff
@@ -0,0 +1,5 @@
+
+> planuj.finance@1.0.0 supabase /home/attila/Projects/Snaha/kalkul.app
+> supabase "db" "diff"
+
+ ELIFECYCLE  Command failed with exit code 1.
```
vojtechsimetka: This looks like an error.

vojtechsimetka (commented):
Code wise looks OK, but I was not able to run it because the migration is wrong.

supabase/migrations/20240906163118_create_portfolio.sql
agazso: Indeed. I used the command `pnpm supabase db diff | pnpm supabase migration new create_portfolio` and it generates a broken output.

I regenerated the migration sql  with using only the `diff` command and manually copying the output to the file.

src/lib/components/add-client.svelte
```diff
@@ -32,13 +33,21 @@
 		imageURI = undefined
 		createClient(client)
 	}
+
+	function cancel(event: Event) {
+		// not sure why this is needed here, but it won't work without it
```
vojtechsimetka: strange indeed
```suggestion
		// FIXME: not sure why this is needed here, but it won't work without it
```

src/lib/components/add-portfolio.svelte
```diff
@@ -0,0 +1,214 @@
...
+			currency,
+			start_date: startDate,
+			end_date: endDate,
+			inflation_rate: inflation,
+		})
+		close()
+	}
+
+	function cancel(event: Event) {
+		// not sure why this is needed here, but it won't work without it
```
vojtechsimetka:
```suggestion
		// FIXME: not sure why this is needed here, but it won't work without it
```

supabase/migrations/20240906163118_create_portfolio.sql
```diff
@@ -0,0 +1,69 @@
...
+grant references on table "public"."portfolio" to "anon";
+
+grant select on table "public"."portfolio" to "anon";
+
+grant trigger on table "public"."portfolio" to "anon";
+
+grant truncate on table "public"."portfolio" to "anon";
+
+grant update on table "public"."portfolio" to "anon";
+
```
vojtechsimetka, line 35: Only authenticated should be able to alter this table.

```suggestion
```

vojtechsimetka: And actually we should have more elaborate checks here (not saying it should be part of this PR)

vojtechsimetka (commented):
I am getting this rather weird bug wen creating client in this branch. It shows all dates as `NaN`. I don't have this issue on `main` but I also don's see anythings in code that could cause this. Please check if you also have this issue.

Screenshot 2024-10-04 at 14 56 01

supabase/migrations/20240906163118_create_portfolio.sql
vojtechsimetka: The `client` table is also vulnerable, so maybe keep it as is and we tackle it once we have all the tables.

vojtechsimetka (commented):
Please enable realtime on the `portfolio` table (annoyingly this is not part of the `supabase db diff`)

```
alter publication supabase_realtime add table public.portfolio;
```

supabase/migrations/20240906163118_create_portfolio.sql
agazso: I opened an issue about this: #225

agazso:
> Please enable realtime on the `portfolio` table (annoyingly this is not part of the `supabase db diff`)
> 
> ```
> alter publication supabase_realtime add table public.portfolio;
> ```

I added it for now, but I am not sure we need realtime at all for this product.

agazso:
> I am getting this rather weird bug wen creating client in this branch. It shows all dates as `NaN`. I don't have this issue on `main` but I also don's see anythings in code that could cause this. Please check if you also have this issue.

This was because I used an intermediate version of the date-input that I patched to accept strings. I fixed this by using the latest date-input from Diete which works with `Date` types.

vojtechsimetka (approved):

PR #228, opened 2024-10-14
feat: create investments

agazso:
This PR adds the investment creation flow.

Notes:
- There is no error checking for the inputs. The best would be to define input components for often used values (number, percentage etc.) and validate when the input value changes.
- There is more and more repetitive code in the Supabase adapter implementation. It would be good to refactor the code to eliminate repetition.
- Removed some more unused code
- Updated Supabase to latest version to see if that fixes the migration issues, but it does not. The command in the README results in a corrupt migration file.
- Also the `typesdb.ts` was corrupt and I had to fix it manually.


supabase/migrations/20241014183452_create_investment.sql
```diff
@@ -0,0 +1,70 @@
...
+grant select on table "public"."investment" to "service_role";
+
+grant trigger on table "public"."investment" to "service_role";
+
+grant truncate on table "public"."investment" to "service_role";
+
+grant update on table "public"."investment" to "service_role";
+
+
+
```
vojtechsimetka: It's missing the realtime so it does not work out of the box because in adapters/supabase/index.ts we subscribe to a realtime channel.
```suggestion

alter publication supabase_realtime add table public.investment;

```

vojtechsimetka (approved):

PR #230, opened 2024-10-22
feat: deposit & withdrawal

agazso:
This PR adds deposit & withdrawals.

Notes: 
- Deposits & withdrawals are called `transaction` in the database, which has a `type` field, indicating if it's a `deposit` or `withdrawal`. Not the best name, but I could not come up with better and didn't want to use `deposit-withdrawal`.
- There is a new `EditTransaction` component for creating and editing transactions. I suggest to follow this pattern in the future and refactor `AddClient`, `AddInvestment` and `AddPortfolio` in a similar way.
- I refactored the `authStore` to be not part of the adapter. Made the `adapter` instance typed as `Adapter` so that to avoid implementation specific details to be used externally. I also refactored the subscriptions in `SupabaseAdapter`.
- Added the `date-fns` library for easy date calculation functionality.

Future TODOs (to be done in different PRs):
- Menu is missing from the investment card (#242)
- `DateInput` is broken in 2 column layout and it's difficult to style because it is implemented internally with an `Input` wrapped in a `div`, but the button is somehow not part of the parent `div`. This needs more work to untangle. (#243)
- Flatten the folder hierarchy. It is getting cumbersome to work with the folders in a 4 level deep hierarchy. It would be better just to use the id on a certain level and recursively look up parents from the stores. (#244)
- Add `user_id` field and policy to tables where it is missing (portfolio, investment) (#245)


vojtechsimetka:
Sadly I can't test this, I keep getting following error even on a new machine with a completely new database and private browsing. I can try and fix the authentication if needed or you can @agazso 

One issue that we have is that when authentication fails, it actually opens the app in a state where it looks like I am signed in as some user. This is linked to #234 .

Screenshot 2024-11-15 at 7 34 24

supabase/migrations/20241108093852_transactions.sql
```diff
@@ -0,0 +1,78 @@
...
+grant truncate on table "public"."transaction" to "service_role";
+
+grant update on table "public"."transaction" to "service_role";
+
+create policy "Enable all for users based on user_id"
+on "public"."transaction"
+as permissive
+for all
+to public
+using ((( SELECT auth.uid() AS uid) = user_id));
```
vojtechsimetka, line 78:
```suggestion
using ((( SELECT auth.uid() AS uid) = user_id));

alter publication supabase_realtime add table public.transaction;
```

vojtechsimetka (approved):

PR #231, opened 2024-10-22
feat: signup flow

hlavkat:
Known issues (to be addressed in separate PR):
- Clicking on password reset will open the app and ask user to input new password. However, if user refreshes the page, it signs them in and the reset password state changes to logged in.


src/lib/adapters/supabase/index.ts
```diff
@@ -273,6 +275,21 @@ export default class Supabase implements Adapter {
 		}
 	}
 
+	async sendResetPasswordLink(email: string) {
+		const { error } = await supabase.auth.resetPasswordForEmail(email)
+		if (error) {
+			console.error('Failed to send reset password link', error)
```
vojtechsimetka, line 281: Does this leak anything? Like if the `email` does not exist?

src/lib/components/login.svelte
```diff
@@ -55,59 +89,179 @@
...
+		<img src="/unboxed-cat.svg" alt="cat" width="320px" />
+		<div class="text">
+			<Typography variant="h4">✅ Email sent!</Typography>
+			<Typography variant="large">
+				{#if window.location.hostname === 'localhost' || window.location.hostname === '127.0.0.1'}
+					Please check your <a href="http://127.0.0.1:64324/m/{email}" target="_blank">inbucket</a> inbox
+					and follow instructions to reset password.
+				{:else}
+					Please check your {email} inbox and follow instructions to reset password.
+				{/if}
```
vojtechsimetka: This text should not be hardcoded, instead it should come from `locales`

src/lib/components/registration.svelte
```diff
@@ -67,61 +98,148 @@
...
+			<Typography variant="h4">✅ Check your email to complete sign up!</Typography>
+			<Typography variant="large">
+				{#if window.location.hostname === 'localhost' || window.location.hostname === '127.0.0.1'}
+					An email with a verification link was sent to your <a
+						href="http://127.0.0.1:64324/m/{user.email}"
+						target="_blank">inbucket</a
+					>.
+				{:else}
+					An email with a verification link was sent to {user.email}
+				{/if}</Typography
```
vojtechsimetka: Also should come from locales

src/lib/components/ui/divider.svelte
```diff
@@ -4,24 +4,38 @@
...
 		margin-right: var(--padding);
 		margin-left: var(--padding);
 		min-width: 1px;
 		height: 100%;
 		min-height: var(--padding);
+		&.noMargin {
+			margin-right: 0;
+			margin-left: 0;
+		}
 	}
```
vojtechsimetka: The question is whether the use with no margin makes sense and it should not be border on some component.

If it is needed though, it should either be overwritten through the `className` or ve could do:
```css
 		margin-right: var(--margin, --padding);
 		margin-left: var(--margin, --padding);
```

and use as:
```html
<Divider --margin={0} />
```

src/routes/+layout.svelte
```diff
@@ -20,12 +27,82 @@
...
+	<div class="intro-screen">
+		<div class="intro-banner">
+			<div class="logo">
+				<img src="/logo.svg" alt="Logo" />
+			</div>
+			<div class="info">
+				<Typography variant="h1">Fingerstache mukbang</Typography>
+				<Typography variant="large"
+					>Viral meggings austin, chicharrones cray vinyl banjo pickled adaptogen bitters affogato
+					cornhole vaporware messenger bag.</Typography
```
vojtechsimetka, line 40: I know that this is in the design, but we should not commit jibberish. @david-gauquelin  can you fix this in design?

supabase/config.toml
```diff
@@ -83,7 +83,7 @@ enabled = true
 enabled = true
 # The base URL of your website. Used as an allow-list for redirects and for constructing URLs used
 # in emails.
-site_url = "http://127.0.0.1:3000"
+site_url = "http://localhost:5173"
 # A list of *exact* URLs that auth providers are permitted to redirect to post authentication.
 additional_redirect_urls = ["https://127.0.0.1:3000"]
```
vojtechsimetka: This should either be removed or also set to `http://localhost:5173`

src/lib/adapters/supabase/index.ts
hlavkat: I think it just checks if the address contains @

src/lib/components/registration.svelte
```diff
@@ -72,6 +72,7 @@
 			formValid = false
 		}
 	})
+	$inspect(user.email)
```
vojtechsimetka: Remove please

vojtechsimetka (approved):

vojtechsimetka (approved):

src/lib/locales/en.json
```diff
@@ -87,6 +87,22 @@
...
+	"forgotPassword": "Forgot password?",
+	"noAccount": "No account yet?",
+	"signUp": "Sign up",
+	"createAccount": "Create account",
+	"haveAccount": "Already have an account?",
+	"resetLink": "Send me a reset link",
+	"goBack": "Go back to ",
+	"updatePassword": "Update password",
+	"forgotPasswordText": "Enter the email associated with your account and you will receive a link to reset your password.",
+	"resetPasswordLocal": "If there’s an account associated {email} you will receive an email to your <a href=\"http://127.0.0.1:64324/m/{email}\" target=\"_blank\">inbucket</a> with instructions to reset your password.",
```
agazso: This does not look correct. It would be better to use a variable (`baseAddress` and define it somewhere):

```suggestion
	"resetPasswordLocal": "If there’s an account associated {email} you will receive an email to your <a href=\"{baseAddress}/m/{email}\" target=\"_blank\">inbucket</a> with instructions to reset your password.",
```

src/lib/locales/en.json
```diff
@@ -87,6 +87,22 @@
...
+	"resetPasswordRemote": "If there’s an account {email} you will receive an email with instructions to reset your password.",
+	"emailSent": "✅ Email sent!",
+	"verificationLinkLocal": "An email with a verification link was sent to your <a href=\"http://127.0.0.1:64324/m/{email}\" target=\"_blank\">inbucket</a>",
```
agazso: Link does not look correct, see my other comment.

src/routes/+layout.svelte
```diff
@@ -20,12 +27,82 @@
...
+			</div>
+			<div class="buttons">
+				<Button onclick={() => (registration = true)}>{$_('signUp')}</Button>
+				<Button onclick={() => (login = true)} variant="secondary"
+					>{$_('login')}<ArrowRight size={24} /></Button
+				>
+			</div>
+		</div>
+		<div class="image">
+			<img src="/intro-image.svg" alt="bird" />
```
agazso: I suggest to not use the Figma exported SVG for illustrations, because it generates quite big files. For example this simple image is more than 100KB. I recommend using PNG instead.

vojtechsimetka: Maybe we should use the originals that @david-gauquelin paid for. I wonder what is their size.

agazso:
It would be good if I could log in by pressing the Enter after the password. Feel free to address this in a separate PR.

agazso:
I am not sure how to test this. It seems that the emails are sent to the inbucket service, but I don't get them. Also it would be good to not use inbucket in the production version.

vojtechsimetka:
> I am not sure how to test this. It seems that the emails are sent to the inbucket service, but I don't get them. Also it would be good to not use inbucket in the production version.

The inbucket is only on localhost (part of supabase docker). But you are right, we could hide it behind env variable.

agazso:
The inbucket email never arrives for me and I cannot login.

I found one issue though: after the login fails there is a warning shown, but that does not disappear when I change my login credentials, instead it just stays there the whole time.

gnome-shell-screenshot-9vmuoz

src/lib/schemas.ts
```diff
@@ -70,8 +70,11 @@ export const detailStoreSchema = portfolioSchema.extend({
 
 export const investmentSchema = z.object({})
 
-export const loginFormSchema = z.object({
-	email: z.string().email({ message: 'Enter a valid email.' }),
+export const emailFormSchema = z.object({
+	email: z.string().email({ message: 'This doesn’t look like a correct email' }),
```
agazso: Should be localised string
```suggestion
	email: z.string().email({ message: $_('This doesn’t look like a correct email' )}),
```

src/lib/schemas.ts
```diff
@@ -70,8 +70,11 @@ export const detailStoreSchema = portfolioSchema.extend({
...
+})
+
+export const loginFormSchema = emailFormSchema.extend({
 	password: z.string().min(6, { message: 'Password must be at least 6 characters long.' }),
```
agazso: Should be localised string:
```suggestion
	password: z.string().min(6, { message: $_('Password must be at least 6 characters long.') }),
```

src/lib/schemas.ts
```diff
@@ -84,7 +87,7 @@ export const registerFormSchema = loginFormSchema
 			ctx.addIssue({
 				code: 'custom',
 				path: ['confirmPassword'],
-				message: 'Passwords must match.',
+				message: 'Passwords don’t match',
```
agazso: Should be localised string:
```suggestion
				message: $_('Passwords don’t match'),
```

agazso (approved):
Please fix the localised strings.

PR #233, opened 2024-10-30
feat: header in layout 

hlavkat:


vojtechsimetka (approved):

vojtechsimetka:
@agazso what do you think about this? It just splits the main `+layout.svelte` into root one which handles login logic and one nested in `(app)` folder which adds header. Then pages that don't have header reset out of the layout.

```
.
└── (app)
    └── client
        ├── [id]
        │   └── portfolio
        │       ├── [portfolio_id]
        │       │   └── investment
        │       │       └── new    // no header
        │       └── new    // no header
        └── new    // no header
```

src/lib/components/ui/dropdown.svelte
agazso, line 1: Why not use the dropdown from the Based project, which is a fork of the one in Diete rather than to create one from scratch?

agazso (approved):

agazso:
> @agazso what do you think about this? It just splits the main `+layout.svelte` into root one which handles login logic and one nested in `(app)` folder which adds header. Then pages that don't have header reset out of the layout.
> 
> ```
> .
> └── (app)
>     └── client
>         ├── [id]
>         │   └── portfolio
>         │       ├── [portfolio_id]
>         │       │   └── investment
>         │       │       └── new    // no header
>         │       └── new    // no header
>         └── new    // no header
> ```

Looks good. BTW I plan to flatten the structure, because with 4 level of hierarchy the urls and folder structrue is getting hard to manage.

PR #235, opened 2024-11-05
ci: create PR previews on digital ocean

vojtechsimetka:


vojtechsimetka:
@agazso so this works very well, but the issue is that each PR spins new instance that costs 5 USD per month. So the more PRs is open and the longer they are open, the more expensive it is. Frankly it will likely not be more expensive than Render (and for sure cheaper than Vercel), but maybe we can optimize a bit.

Few ideas:
- we add button that trigger spinning up the instance when needed. So it is not automatic on every PR but it need to be triggered on/off. We can also disable it for specific branches like the release branch.
- we could have a keyword that needs to be in the PR title/comment that triggers the PR preview.
- we keep Render but we create new account specific for Kalkul that should suffice with free minutes. This is better preview experience but worse devops experience as we need to keep env variables etc in two places.

agazso:
> @agazso so this works very well, but the issue is that each PR spins new instance that costs 5 USD per month. So the more PRs is open and the longer they are open, the more expensive it is. Frankly it will likely not be more expensive than Render (and for sure cheaper than Vercel), but maybe we can optimize a bit.
> 
> Few ideas:
> 
>     * we add button that trigger spinning up the instance when needed. So it is not automatic on every PR but it need to be triggered on/off. We can also disable it for specific branches like the release branch.
> 
>     * we could have a keyword that needs to be in the PR title/comment that triggers the PR preview.
> 
>     * we keep Render but we create new account specific for Kalkul that should suffice with free minutes. This is better preview experience but worse devops experience as we need to keep env variables etc in two places.

The button sounds nice.

PR #237, opened 2024-11-06
feat: account settings

hlavkat:


src/routes/account/+page.svelte
```diff
@@ -0,0 +1,143 @@
...
+	import Button from '$lib/components/ui/button.svelte'
+	import Input from '$lib/components/ui/input/input.svelte'
+	import Option from '$lib/components/ui/select/option.svelte'
+	import Select from '$lib/components/ui/select/select.svelte'
+	import Typography from '$lib/components/ui/typography.svelte'
+	import { Checkmark } from 'carbon-icons-svelte'
+	import { locale, locales, _ } from 'svelte-i18n'
+
+	let emailChange = $state(false)
+	let passwordChange = $state(false)
```
vojtechsimetka: Please instead create pages for this:

```
/account/change-email
/account/change-password
```

src/lib/components/ui/select/select.svelte
```diff
@@ -13,7 +13,7 @@
 		label?: string
 		labelFor?: string
 		dimension?: Dimension
-		value?: string
+		value?: string | null
```
vojtechsimetka: this should probably be propagated to Diete

vojtechsimetka (changes requested):

src/lib/components/ui/select/select.svelte
agazso: Please don't use `null`, use `undefined` instead

@vojtechsimetka can we make an eslint rule for this?

src/routes/account/+page.svelte
```diff
@@ -0,0 +1,81 @@
...
+	import { locale, locales, _ } from 'svelte-i18n'
+
+	const languageName: Record<string, string> = {
+		en: 'English',
+		cs: 'Čeština',
+	}
+
+	$effect(() => {
+		if ($locale && !$locales.includes($locale)) {
+			$locale = $locale.split('-')[0]
```
agazso, line 19: It does not seem to be a good idea to overwrite a variable that is imported from a library. It would be better to use a local `$state` or `$derived` instead.

agazso (approved):
Generally works well.

I found minor issues which can be addressed in later PRs:
- The email address field is not `disabled` (because that would change the color). I see that you used the `readonly` property. However this makes the email field to be focused when you click on it and selects all the text. You may want to try to use the `inert` property (see also https://github.com/snaha/diete.design/issues/390)
- If the password is incorrect on the `Change password` page the error is `Invalid credentials`, which I guess was copied from the login page. It is technically not incorrect, but still feels strange here, where it would be better to say `Invalid password`. Please note that on the login page the error is different because we don't want to reveal extra information about the existence of emails or the passwords being incorrect.

agazso:
After testing a bit more it turns out that the email changing functionality is not working correctly, I opened an [issue about it](#259).

PR #247, opened 2024-11-19
feat: transaction card menu

### LINKED ISSUE #242: Transaction card menu

agazso:
This PR adds the transaction card menu. Closes #242 

Notes:
- It was not trivial to make the dropdown work well on an element which is itself clickable. `stopPropagation()` cannot be used because it makes the other dropdowns avoid detecting the click event, therefore they would not close. The solution is to use the `defaultPrevented()` and ignore the click on the transaction card if `preventedDefault` is set to `true`.
- The dropdown is cleaned up a bit, I removed one `div` and some non-functional CSS properties.
- I added a default `notImplemented` function implementation, but did not change existing ones to use that.


src/lib/components/transaction-card.svelte
```diff
@@ -48,7 +86,31 @@
...
+				>
+				<Button variant="ghost" dimension="compact" leftAlign onclick={duplicateTransaction}
+					><Copy size={24} />{transaction.type === 'deposit'
+						? $_('Duplicate deposit')
+						: $_('Duplicate withdrawal')}</Button
+				>
+				<Button variant="ghost" dimension="compact" leftAlign onclick={deleteTransaction}
+					><TrashCan size={24} />{transaction.type === 'deposit'
+						? $_('Delete deposit')
+						: $_('Delete withdrawal')}</Button
```
vojtechsimetka, line 110: I don't see were are these translations defined.

vojtechsimetka (approved):
Works fine, not sure about the text if it is defined or just hardcoded. I would love to have it come the localization json

src/lib/components/transaction-card.svelte
agazso: Added the definitions for the files that were affected by this commit. Also made sure that if there are ternary expressions involved then the translation is the result of the expression so that it's easier for search for missing translations with a potential tool.

PR #254, opened 2024-11-25
feat: share

### LINKED ISSUE #250: Share link

agazso:
This PR implements the portfolio sharing with a link feature. Closes #250

Notes:
- Finding a way to get data from the database with Row Level Security turned on is not trivial. My original idea was using views, but it turns out that is not possible for several reasons. I managed to get it work with SQL functions, because it is possible to use them with `SECURITY DEFINER` which means that the function runs with the same privilege as the user who created the function. The issue is that it requires defining types for the return type in SQL, which is a duplicate of the data types of the tables, so we will have to pay attention to it anytime we change the type definitions for tables `client`, `portfolio`, `investment` or `transaction`. There is also no runtime typechecking when the data is returned from the database so this is fragile. 
- The alternative to using SQL functions could be a server endpoint which connects to the database with a high-privilege user, makes the queries and return the data as JSON.
- The confirmation modal when turning off link sharing is missing. #255 
- The actual graphs are missing because they are not yet implemented. When they are done, they need to be added to the readonly view as well.
- I added the `/view` endpoint outside the app, but unfortunately the login screen still shows up because the check for user authentication happens in `src/routes/+layout.svelte`. This should be done to the `(app)` level instead. #256
- When trying to use WebCrypto in the browser Svelte throws an error and I did not find a solution for that. There is a workaround which generates a non-cryptographic random which is good enough considering the length of the string and because we don't use the random for cryptography, just as a random identifier. Still it would be nice to be able to use the  WebCrypto API.


src/lib/components/add-portfolio.svelte
```diff
@@ -45,6 +45,7 @@
 			start_date: startDate.toDateString(),
 			end_date: endDate.toDateString(),
 			inflation_rate: inflation,
+			link: null,
```
vojtechsimetka, line 48: To stay consistent :)
```suggestion
			link: undefined,
```

vojtechsimetka (commented):
I get the following issue when trying to use the migration locally. Do I need to enable something?

`supabase db reset`

```
WARN: no SMS provider is enabled. Disabling phone login
Resetting local database...
Recreating database...
Setting up initial schema...
Seeding globals from roles.sql...
Applying migration 20240826115858_registration_and_client_table.sql...
Applying migration 20240826125306_enabled_realtime_on_client.sql...
Applying migration 20240906163118_create_portfolio.sql...
Applying migration 20241014183452_create_investment.sql...
Applying migration 20241108093852_transactions.sql...
Applying migration 20241125152233_share.sql...
ERROR: type "client_view_type" does not exist (SQLSTATE 42704)                         
At statement 4: CREATE OR REPLACE FUNCTION public.client_readonly_view(link_id text)   
 RETURNS SETOF client_view_type                                                        
 LANGUAGE sql                                                                          
 SECURITY DEFINER                                                                      
AS $function$                                                                          
    SELECT client.id, client.name, client.birth_date, client.advisor, client.created_at
    FROM client                                                                        
    JOIN portfolio ON portfolio.client=client.id                                       
    WHERE portfolio.link=link_id                                                       
  $function$                                                                           
Try rerunning the command with --debug to troubleshoot the error.
```

agazso:
> I get the following issue when trying to use the migration locally. Do I need to enable something?
> 
> `supabase db reset`
> 
> ```
> WARN: no SMS provider is enabled. Disabling phone login
> Resetting local database...
> Recreating database...
> Setting up initial schema...
> Seeding globals from roles.sql...
> Applying migration 20240826115858_registration_and_client_table.sql...
> Applying migration 20240826125306_enabled_realtime_on_client.sql...
> Applying migration 20240906163118_create_portfolio.sql...
> Applying migration 20241014183452_create_investment.sql...
> Applying migration 20241108093852_transactions.sql...
> Applying migration 20241125152233_share.sql...
> ERROR: type "client_view_type" does not exist (SQLSTATE 42704)                         
> At statement 4: CREATE OR REPLACE FUNCTION public.client_readonly_view(link_id text)   
>  RETURNS SETOF client_view_type                                                        
>  LANGUAGE sql                                                                          
>  SECURITY DEFINER                                                                      
> AS $function$                                                                          
>     SELECT client.id, client.name, client.birth_date, client.advisor, client.created_at
>     FROM client                                                                        
>     JOIN portfolio ON portfolio.client=client.id                                       
>     WHERE portfolio.link=link_id                                                       
>   $function$                                                                           
> Try rerunning the command with --debug to troubleshoot the error.
> ```

I manually edited the generated migration SQL file so that the type declaration is before the function definition which uses it. Please try again. 

If it works it means that this is not handled automatically by supabase and we will have to pay attention next time if we use a similar construct.

src/lib/components/add-portfolio.svelte
agazso: I would be happy not using `null`, but unfortunately in SQL there is a `NULL` value indicating a missing value and the Supabase type generator converts that to Typescript `null` value.

See here:
https://github.com/snaha/kalkul/pull/254/files#diff-bffd5e4300c7d550a0d30d4c443baf638b9e493bc96278ef30f71ae40f9b74d7R151

vojtechsimetka (approved):
Nice job, works fine.

About WebCrypto API. It works in Svelte, but only browser side. So you first need to check if `browser` is true to use it.

agazso:
> About WebCrypto API. It works in Svelte, but only browser side. So you first need to check if `browser` is true to use it.

I tried that and it did not work unfortunately, but I did not spend too much time on it. Feel free to tackle this.

PR #260, opened 2024-12-04
refactor: layout

### LINKED ISSUE #246: Improve authentication
Right now, on reload the page loads into unauthenticated state first before logging in. It should probably be in some loading state first.

### LINKED ISSUE #256: Move user authentication check to `(app)` level
The login screen shows up for unauthenticated pages, because the check for user authentication happens in `src/routes/+layout.svelte`. This should be done to the `(app)` level instead.

agazso:
This PR refactors the layout so that the app is authenticated and the header is correctly applied. 

Fixes #256 
Fixes #246 

Notes:
- The authentication is moved to the `(app)` level `+layout.svelte`. That means that all the app level endpoints are authenticated.
- The header is moved to a separate `Header` component and it is applied manually to the `(app)` level `+page.svelte` and  automatically to all routes under `(app)/client`. When a page does not require a header it can be placed either at the `(app)` level or can [break out the layout](https://svelte.dev/docs/kit/advanced-routing#Advanced-layouts-Breaking-out-of-layouts) either at layout or page level.
- The `/view/[id]` route is now outside the `(app)` so it is no longer authenticated.
- There is a workaround (`authStore.signingIn`) to avoid displaying the intro page when reloading the page when the user is already authenticated. The root cause of the issue is that an `INITIAL_SESSION` event may arrive after a `SIGNED_IN` event and if that happens it needs to be stored somewhere, hence the workaround.


vojtechsimetka (approved):
Nice!

About 5% of refreshes it puts me on the login screen. I am fine with merging this and tackling that separately. (see the start and end of attached video). There is already somewhat of an issue about it #246

https://github.com/user-attachments/assets/2ddaf430-eeea-4d6d-834e-6deff1fafb60

agazso:
> Nice!
> 
> About 5% of refreshes it puts me on the login screen. I am fine with merging this and tackling that separately. (see the start and end of attached video). There is already somewhat of an issue about it #246
> Screen.Recording.2024-12-04.at.16.54.24.mov

Interesting, I cannot reproduce it. I opened an issue about it, I have an idea what might cause it: #261 
I meant to address #246 too.

PR #262, opened 2024-12-06
feat: client menu

### LINKED ISSUE #253: Edit client

agazso:
This PR adds the client drop-down menu, client editing and client deletion features. 

Closes #253 

Notes:
- Added the `Modal` component
- Added `List` and `ListItem` components to be used with the `Dropdown` component (which is updated)
- Added a `Fullscreen` component for transitional screens


src/routes/(app)/+page.svelte
```diff
@@ -2,23 +2,69 @@
...
+		{/snippet}
+		<List>
+			<ListItem onclick={() => goto(routes.CLIENT(clientId))}
+				><FolderShared size={24} />{$_('View portfolios')}</ListItem
+			>
+			<ListItem onclick={() => goto(routes.EDIT_CLIENT(clientId))}
+				><UserProfile size={24} />{$_('Edit client details')}</ListItem
+			>
+			<ListItem onclick={() => confirmDeleteClient(clientId)}
+				><TrashCan size={24} />{$_('Delete client')}</ListItem
```
vojtechsimetka, line 62: These should come from `locales/en.json` but I wonder if for time reasons we should not just hardcore these and then do the translations when reviewing the finished app.

src/lib/components/delete-client-modal.svelte
```diff
@@ -0,0 +1,68 @@
...
+<Modal {oncancel} bind:open {...restProps}>
+	<section class="dialog">
+		<header class="horizontal">
+			<Typography variant="h5">{$_('Delete client?')}</Typography>
+			<div class="grower"></div>
+			<Button variant="ghost" dimension="compact" onclick={oncancel}><Close size={24} /></Button>
+		</header>
+
+		<Typography
+			>This client and all the portfolios it contains will be deleted permanently. There’s no undo.</Typography
```
vojtechsimetka: Hardcoded text that should come from locales

vojtechsimetka (approved):

src/routes/(app)/+page.svelte
agazso: Yeah, it's a bit annoying, because the workflow of adding a text requires switching between two files and copy-pasting the text, which when there are multiple strings involved breaks the coding flow and is error prone (see your other comment, where I added the text to the json, but forgot to add the `$_() function`.)

I could imagine using some tooling to detect if there are non-localized strings and missing localizations from the json. But without that tooling it is easy to forget about it or make mistakes, and we are still in the discovery phase where these text may change, so I could also imagine just ignoring it before the MVP and update them at one step before the launch.

PR #263, opened 2024-12-06
feat: portfolio editing

### LINKED ISSUE #249: Portfolio settings
Also add to portfolio list and portfolio view button

### LINKED ISSUE #251: Edit portfolio
Currently only the add portfolio functionality is implemented.

agazso:
This PR adds the portfolio editing options, both to the dropdown list on the client page and the portfolio page.

Fixes #249 
Fixes #251

Notes
- During testing I found that the inflation cannot be changed from the default value, I opened an issue about it: #264 
- The naming conventions are not clearly defined, and as a result I had some surprises during testing. Unfortunately the compiler does not always infer the correct type and cannot help in all cases. #265
- In order to delete a record in SQL which has references from other tables, the [table definition needs to be changed]( https://stackoverflow.com/questions/69251891/delete-associated-records-in-supabase). Eventually I chose to implement the logic in Javascript by recursively deleting the references, because we also need to support duplicating certain records with a similar logic.


vojtechsimetka (approved):
I would love for the delete to be on the DB side and not in JS. But this could be done in a separate PR.

PR #269, opened 2024-12-13
feat: portfolio and investment graph

vojtechsimetka:
TODO:
- [x] add graph to portfolio page
- [x] add graph to share page
- [x] polish investment graph
- [x] only show deposits & withdrawals for current investment
- [x] generalize the graph view
- [x] dynamically calculate amount of values for the graph
- [x] handle no data
- [x] show data points
- [x] generalize the calculations for portfolio
- [x] gain calculation (no fees)

Not part of this PR (will create issues for this)
- [ ] calculation with fees
- [ ] tests for the calculation
- [ ] detail graphs under the main graph
- [ ] remove the 100% height on graph view (this should be changed on the sidebar as well)
- [ ] graphs are sometimes not getting smaller on Chrome


agazso:
There is another issue, the portfolio above has 250% inflation but that seems to be ignored as well. Also I cannot change it, although it has been fixed in #275

agazso (approved):
Works well!

PR #287, opened 2024-12-17
fix: date input

agazso:
This PR adds minor fixes to the `DateInput` component.

- Fixes the button border when the input is `solid`
- Removes unnecessary padding from the `input` element so that the sizing is less off, but it's still not perfect

The end result is that on Chrome the `DateInput` is one pixel longer than it should be, on Firefox it still displays the default calendar icon and that makes the whole layout broken.

There are other issues:
- When the button is selected with the keyboard the calendar icon moves one pixel to the right. 
- The `DateInput` is taller than the text input field on Chrome (but not on Firefox)

I see two options improving this:
1. Complete rewrite of the `DateInput` component which would not use the underlying `input` with `type="date"` but rather implement the component with either 3 input components or generating an input dynamically depending if you select the year, month or day. With this we could have complete control over how the component looks and how it is laid out. The downside is that this would require significant work.
2. Use just a simple date picker instead. This is what Shadcn provides.


vojtechsimetka (approved):

vojtechsimetka:
I have slight preference from option 2, but option 1 is also fine.

PR #297, opened 2024-12-19
feat: add Reddit Sans font

agazso:
This PR adds the `Reddit Sans` font as the default font. It has the `font-feature-setting: 'tnum'` setting, which means that it [displays numbers as a monospace font](https://otf.thien.do/tnum). Therefore the `sans` font can be used everywhere and there is no need to use a monospace font.


vojtechsimetka (approved):

david-gauquelin:
In the currently deployed version on kalkul.app the numbers are not using the monospace style. Should I open an issue about this?

Screenshot 2025-01-16 at 9 04 04

PR #301, opened 2024-12-23
feat: new style

agazso:
This PR adds the updated style.

Personally I think the paddings are too big and the information density is a bit low. It does not help that the planned headers are fixed on the top, because they take up a lot of valuable vertical space.

My suggestion is to adjust the paddings, mainly the vertical, but as a consequence it will require changing some horizontal paddings as well. Maybe only make it bigger after a certain screen size. Also reconsider the use of default size buttons in the header.

An alternative could be to introduce a setting for information density. Professional tools often have it either as options (could be compact or default as a start, like in Gmail or Thunderbird) or with zoom level (for example in Google Spreadsheet).


vojtechsimetka (approved):
Very nice!

src/routes/(app)/+layout.svelte
```diff
@@ -44,7 +44,7 @@
 			</div>
 		</div>
 		<div class="image">
-			<img src="/intro-image.png" alt="bird" width="100%" />
```
vojtechsimetka, line 47: Maybe we should erase the `intro-image.png`?

PR #343, opened 2025-01-09
fix: disabled input on account page

### LINKED ISSUE #315: Email input should be disabled on Account settings page
Screenshot 2025-01-08 at 9 39 06

see the disabled design for inputs in Diete
https://www.figma.com/design/aJyCbDzi48QRb33vSIKt0J/DieteUI_KALKUL_02?node-id=23-966&t=WDKuJG5zT2aElWZM-1

hlavkat:
resolved #315


agazso (changes requested):
This is not how the design looks.

The solution is to change the `Input` component disabled style so that the text color does not change (i.e. `ultra-high`) but the border color is changed to `low`.

agazso (approved):

Issue #296, opened 2024-12-19
The deposit/withdrawal tooltip should display the name


agazso:
![gnome-shell-screenshot-anpqui](https://github.com/user-attachments/assets/61b762f8-d1f5-4bd5-83a1-38f2b073e6c5)

agazso:
Closed as completed

PR #356, opened 2025-01-10
feat: adjust with inflation

### LINKED ISSUE #352: Implement inflation toggle

### LINKED ISSUE #361: APY calculation is incorrect
I have an investment with 5% APY with an initial deposit of `1000` EUR. There is also a recurring deposit with `0.00001` because of issue #360, but that should not matter much. I would expect that after a year the value of my investment is at least `1050`, instead I have `1026`.

gnome-shell-screenshot-esb02g

vojtechsimetka:
resolves #352
resolves #361

Not part of this PR:
- Showing non-inflation calculated values in the tooltip for both graphs
- Omitting 0 values in the tooltip for deposit/withdrawal graph
- Separating withdrawals and deposits in the tooltip for deposit/withdrawal graph


agazso (commented):
The good news is that this seems to fix the issue in #361. The bad news is that changing the inflation toggle does not seem to be doing anything: with 2.25% inflation for 30 years the calculated value is the same with inflation as without it.

gnome-shell-screenshot-86m0h6

gnome-shell-screenshot-hjqcux

vojtechsimetka:
Yeah the problem in #361 is actually that by default inflation was enabled for the portfolio chart.

I'll fix the portfolio calculation. Good bug find.

vojtechsimetka:
@agazso please check if this is fixed for you.

Screenshot 2025-01-13 at 20 55 15

Screenshot 2025-01-13 at 20 55 23

vojtechsimetka:
After some more testing, it seems that this somehow depends on the date of the investment and for date `now()` the inflation toggle does nothing.

src/lib/calc.ts
```diff
@@ -182,23 +182,70 @@ export function getGraphData(
...
 	)
 
+	const inflationPerPeriod = getInflationAdjustment(portfolio.inflation_rate, frequency, count)
+
+	const graphInflationDeposits: number[] = []
+	const graphInflationWithdrawals: number[] = []
+	const graphInflationInvestmentValue: number[] = []
+
+	for (let i = 0; i < graphLabels.length; i++) {
+		const inflation = inflationPerPeriod.pow(i)
```
vojtechsimetka, line 194: Technically this can be optimised to not use `pow` here but instead remember this value between iterations and use just multiplication. However, I think this is not too much of a computational overhaul.

agazso (approved):
Nice!

It would be good if the labels in the deposit/withdrawal chart tooltip displayed the name of the deposit/withdrawal and not the name of the investment (Edit: I found the relevant issue: #296).

vojtechsimetka:
Yeah I'll tackle that next.

PR #358, opened 2025-01-10
feat: date input

### LINKED ISSUE #309: Add client not accepting default birthdate
In the add client page, if I don't change the default date for birthdate it won't let me create the client.
We should either:
- communicate this to the user somehow (maybe with an empty state for the date input, just showing a placeholder YYYY/MM/DD)
- or allow to create with the default selected date

The first option is better I think

agazso:
This PR implements the date input component according to current design.

Fixes #309 

Notes:
- If the date is invalid the passed in bindable `value` prop is undefined
- The placeholders for the inputs and the error messages can be passed in as props, so the component can be localized
- The input tab focus automatically changes when the input is filled on the year and month inputs
- The year/month/day can also be changed with pressing the up/down key on the keyboard
- The calendar is now a separate component and is also updated according to the latest design
- The calendar button has a tabindex `-1` so that it cannot be selected from the keyboard. However when the calendar is visible after opening with clicking on the button, then the keyboard navigation works on the calendar component.

Known issues and possible improvements:
- The small calendar width changes when switching between the date picker and year picker. Most likely this is because the design uses 2 letter month names, whereas the code uses the builtin `Intl.DateTimeFormat(undefined, { weekday: 'short' })` format which uses 3 letter month names, and therefore at small size there is a visible difference.
- It's possible to add further improvements for keyboard navigation: pressing right when at the right edge of the year could advance towards the month input, and similarly pressing left when at the left edge of the month input could go back to the end of the year input. Backward navigation with the backspace key is also a possible improvement.
- Another improvement could be to not let enter invalid values at all. For example in the month field, when it's empty only allow `0`..`9`, and only let a second character if the first one is either `0` or `1`. Similarly for the day, but there it's also possible to add a further constraint to only allow valid days in the month. This would require checking the year as well in the case of a leap year when the date is Februrary 29th.


src/lib/components/ui/input/date/input.svelte
```diff
@@ -0,0 +1,368 @@
...
+	}
+
+	function onMonthBlur() {
+		if (!validateMonth()) {
+			error = errorMessages.invalidMonth
+			return
+		}
+
+		if (month.length === 1) {
+			month = '0' + month
```
vojtechsimetka:
```suggestion
			month = `0${month}`
```

Or

```suggestion
			month = month.padStart(2, '0')
```

src/lib/components/ui/input/date/input.svelte
```diff
@@ -0,0 +1,368 @@
...
+	}
+
+	function onDayBlur() {
+		if (!validateDay()) {
+			error = errorMessages.invalidDay
+			return
+		}
+
+		if (day.length === 1) {
+			day = '0' + day
```
vojtechsimetka:
```suggestion
			day = `0${day}`
```

Or

```suggestion
			day = day.padStart(2, '0')
```

src/lib/components/ui/input/date/input.svelte
```diff
@@ -0,0 +1,368 @@
...
+
+	function validateDay(daysOfMonth = 31) {
+		if (!validateNumber(day)) {
+			return false
+		}
+
+		const numValue = parseInt(day, 10)
+		if (!validateRange(numValue, 1, daysOfMonth)) {
+			return false
+		}
```
vojtechsimetka, line 141: It looks like you never set the `daysOfMonth` to anything else than 31. Maybe it can be completely omitted because even for e.g. February you count with 31 days and find out it is wrong date in the `setDateValue` function below.

vojtechsimetka (commented):
Nice. I really like the UX thoughts like switching focus from year input to month input when user types in 4 digits or adding `0` in front of single digit month/day.

Few things i noticed:
- if the date is in error state, I can still create a client. As I create the client, the date returns to the latest valid date.
- navigating with keyboard jumps from the `date` input to the URL. I would expect it to go to `Create client` and `Cancel`. I think that it is correct that the calendar dropdown is not keyboard navigable

https://github.com/user-attachments/assets/400d49af-17c0-40fd-9297-bf05e6d440b2

src/lib/components/ui/input/date/input.svelte
agazso: There is a different error message if the day is outside of the range of 1-31 (`Invalid day`).

src/lib/components/avatar.svelte
```diff
@@ -11,7 +11,7 @@
 
 	let { name, birthDate, imageURI, size = 40 }: Props = $props()
 
-	const seed = $derived(name + ' ' + birthDate.getDate())
+	const seed = $derived(name + ' ' + Number(birthDate))
```
vojtechsimetka:
```suggestion
	const seed = $derived(`${name} ${Number(birthDate)}`)
```

vojtechsimetka (approved):
Nice. There is only one issue with the calendar and that is that some months need more rows. E.g. May 2016 has 31 days but it only shows 29 days.
Screenshot 2025-01-15 at 12 04 52

(And here is June)
Screenshot 2025-01-15 at 12 04 56

PR #362, opened 2025-01-13
feat: table row hover state

### LINKED ISSUE #320: Add hover states for table rows
It would be good to enhance a bit the UX experience of the tables

hlavkat:
resolved #320


agazso (changes requested):
It's better to not define a new variable when it's enough to just apply opacity on a color.

Here is how you can do it: https://stackoverflow.com/questions/40010597/how-do-i-apply-opacity-to-a-css-color-variable

vojtechsimetka (approved):

agazso (approved):

PR #363, opened 2025-01-13
fix: default birtdate on add client page

### LINKED ISSUE #309: Add client not accepting default birthdate

hlavkat:
resolved #309


vojtechsimetka:
Lets build this on top of #358

agazso:
This was fixed in #358

PR #365, opened 2025-01-14
fix: implementation search function - filter by name

### LINKED ISSUE #312: Search is not working
Screenshot 2025-01-08 at 9 35 13

hlavkat:
resolved #312


src/routes/(app)/+page.svelte
```diff
@@ -26,6 +26,8 @@
 
 	let showConfirmModal = $state(false)
 	let clientToBeDeleted: number | undefined = $state()
+	let searchQuery = $state('')
+	let filteredClient = $derived(searchByName())
```
vojtechsimetka:
```suggestion
	let filteredClient = $derived(searchByName(searchQuery))
```

src/routes/(app)/+page.svelte
```diff
@@ -43,6 +45,15 @@
 			showConfirmModal = false
 		}
 	}
+	function searchByName() {
```
vojtechsimetka:
```suggestion
	function searchByName(searchQuery: string) {
```

vojtechsimetka (commented):
It is generally an anti-pattern to have state variables in functions. These should be passed in as arguments.

vojtechsimetka (approved):

PR #380, opened 2025-01-20
feat: date picker scroll improvements

agazso:
Small improvement to scroll the screen when the date picker appears so that it is fully visible. It also remembers the original scroll position and scrolls back if the date picker button is not visible when closing the calendar.

The ideal would be to check if there is space above the button and put the calendar there to avoid unnecessary scrolling. That can be done later as another improvement.


src/lib/components/ui/input/date/calendar.svelte
```diff
@@ -320,9 +348,24 @@
...
 		}
 	}
+	@keyframes fadeInFromNone {
+		0% {
+			opacity: 0;
+		}
+
+		1% {
+			opacity: 0;
+		}
```
vojtechsimetka: Is this for some further easing in or left here by accident?

vojtechsimetka (approved):

src/lib/components/ui/input/date/calendar.svelte
agazso: It was left from an earlier version, removed it.

PR #385, opened 2025-01-20
fix: correctly calculating total deposits and withdrawals on investment card

### LINKED ISSUE #292: Total deposit is incorrect on investment card
It only adds the initial deposit values, does not sum over the recurring deposits.

hlavkat:
resolved #292


agazso (approved):
This works well, but the code is duplicated. It would be good to extract this to a function that can be used at both places and also on other screens.

vojtechsimetka (changes requested):
- as Attila mentioned, please extract this to a reusable function
- we have also daily, weekly and yearly deposits/withdrawals. Also you can have these deposits every x repeats (e.g. every 2 weeks). Please ensure all these options are covered correctly.

hlavkat:
@vojtechsimetka @agazso 
I discovered a problem when I have a recurring deposit of, for example, 1000 euros every day for 5 days... the function counts 5000, but the graph shows 6000. How should it be calculated as 24h or 24h + the current day?

vojtechsimetka:
@hlavkat yeah it should be inclusive. So if I have recurring deposit that starts today and ends today, it should count as 1. If I have one from today to tomorrow that is 2 deposits total.

src/lib/calc.ts
```diff
@@ -319,3 +321,42 @@ function getInvestmentValues(
...
+				transaction.end_date || transaction.date,
+				transaction.repeat_unit || 'month',
+				transaction.repeat || 1,
+			)
+
+			return prev + transaction.amount * periods
+		}, 0)
+}
+
+// Funkce pro výpočet počtu opakování v zadaném období
```
vojtechsimetka:
```suggestion
```

src/lib/calc.ts
```diff
@@ -319,3 +321,42 @@ function getInvestmentValues(
...
+export function numOccurrences(
+	start: string,
+	end: string,
+	repeatUnit: string,
```
vojtechsimetka, line 342: I think we have a type for this, not sure this should be just string (but if it is typed as string in the `Transaction` type then leave it like this)

vojtechsimetka (changes requested):
Please make the end date inclusive. Outside of that this is fine.

src/lib/calc.ts
hlavkat: Yes in Transaction is a string

src/lib/calc.ts
```diff
@@ -349,13 +347,13 @@ export function numOccurrences(
...
+			return Math.max(differenceInDays(endDate, startDate) + 1, 1) / repeat
 		case 'week':
-			return Math.max(differenceInWeeks(endDate, startDate), 1) / repeat
+			return Math.max(differenceInWeeks(endDate, startDate) + 1, 1) / repeat
 		case 'month':
-			return Math.max(differenceInMonths(endDate, startDate), 1) / repeat
+			return Math.max(differenceInMonths(endDate, startDate) + 1, 1) / repeat
 		case 'year':
-			return Math.max(differenceInYears(endDate, startDate), 1) / repeat
+			return Math.max(differenceInYears(endDate, startDate) + 1, 1) / repeat
```
vojtechsimetka: Does this still need to be in `Math.max`? :)

src/lib/calc.ts
agazso: It is true that the `type` field of the `Transaction` type is a string, because it is coming from the Postgres definition. However there is `TransactionType` defined in `types.ts` as `'deposit' | 'withdrawal'`, exactly for this kind of use-cases.

src/lib/calc.ts
```diff
@@ -319,3 +321,40 @@ function getInvestmentValues(
...
+				transaction.date,
+				transaction.end_date || transaction.date,
+				transaction.repeat_unit || 'month',
+				transaction.repeat || 1,
+			)
+
+			return prev + transaction.amount * periods
+		}, 0)
+}
+export function numOccurrences(
```
agazso, line 339: It would be better to call this function `calculateNumOccurrences` because then the name tells what the function does and it also won't clash with a variable called `numOccurrences`. The reason this did not cause a problem is that the variable has a typo in its name (see `numOccurrences` vs. `numOccurences`).

PR #388, opened 2025-01-20
fix: both graphs have same gridlines and labels spacing

### LINKED ISSUE #291: Top chart units are different from bottom chart
It would be good to have the same unit for both charts so that they can be easily compared visually.

ksnip_20241219-220321

vojtechsimetka:
resolves #291

This is now also configurable so we can decide the spacing of labels and gridlines

Screenshot 2025-01-20 at 23 11 14


agazso (approved):
Works well.

Now that I tested it I realized that it is confusing that the points on a 20 year long plan are every 3 months. This is especially confusing on the deposit graph, where you want to see how much it is deposited or withdrawn for a year and you have to multiply it in head. 

Even more confusing that sometimes the points are every 6 months and it is not clear from the label (e.g. `2033-6`) if it's 3 or 6 month, so you have to scan manually and then do the multiplication in your head accordingly.

I would either:
1. Use only 1/year points, then you don't have to multiply anything, just read the tooltip values as is
2. Make this a setting

PR #400, opened 2025-01-24
feat: new design of portfolio according to kalkul05

hlavkat:


src/lib/stores/investment.svelte.ts
```diff
@@ -22,6 +25,12 @@ export function withInvestmentStore(): InvestmentStore {
 		get loading() {
 			return loading
 		},
+		get hiddenInvestments() {
+			return hiddenInvestments
+		},
+		set hiddenInvestments(newValue) {
+			hiddenInvestments = newValue
+		},
```
vojtechsimetka: You don't need the setter.

src/lib/stores/investment.svelte.ts
```diff
@@ -1,14 +1,17 @@
...
 	data: Investment[]
+	hiddenInvestments: SvelteMap<number, boolean>
 	loading: boolean
 	reset: () => void
 	filter: (portfolioId: number) => Investment[]
 }
 
 export function withInvestmentStore(): InvestmentStore {
 	let data = $state<Investment[]>([])
+	let hiddenInvestments = $state<SvelteMap<number, boolean>>(new SvelteMap())
```
vojtechsimetka: This should not be in the store, it should be just in the +page. Make so that it can be easily reused in the `/investment` and `/view` pages.

The store should hold just data from the DB. Right now the nice thing is that as I exit and enter the portfolio, it remembers hidden investments. But this is most likely not what I want because it makes it easy to forget that some investment may be hidden. At the same time, in one portfolio I can affect other portfolios visibility (by focusing and unfocusing on investment).

src/routes/(app)/client/[id]/portfolio/[portfolio_id]/edit-investment/[investment_id]/+page@.svelte
```diff
@@ -17,6 +17,7 @@
 	function close() {
 		history.back()
 	}
+	$inspect(investment)
```
Comment (vojtechsimetka) [inline src/routes/(app)/client/[id]/portfolio/[portfolio_id]/edit-investment/[investment_id]/+page@.svelte]:
```suggestion
```

src/lib/components/edit-transaction.svelte
```diff
@@ -79,41 +79,43 @@
...
+				user_id: authStore.user?.id,
+				investment_id: investment.id,
+				type: transactionType,
+				amount,
+				label,
+				date: date.toDateString(),
+				repeat: isRecurring ? repeat : null,
+				repeat_unit: isRecurring ? repeatUnit : null,
+				end_date: isRecurring ? endDate.toDateString() : null,
+			})
```
vojtechsimetka: Let's add the check from line `82` (`if (investment)`) to line `79` (`if (!authStore.user || !iinvestment)`)

src/lib/components/edit-transaction.svelte
```diff
@@ -79,41 +79,43 @@
...
 		close()
 	}
 
 	async function editTransaction() {
-		if (!transaction.id) {
+		if (!transaction) {
```
vojtechsimetka, line 97:
```suggestion
		if (!transaction || !investment) {
```

And remove it from below

src/lib/components/edit-transaction.svelte
```diff
@@ -33,15 +32,16 @@
 	import FlexItem from './ui/flex-item.svelte'
 
 	type Props = {
-		investment: Investment
+		investment?: Investment
```
vojtechsimetka: This should not be optional

src/lib/components/edit-transaction.svelte
```diff
@@ -56,21 +56,21 @@
 	const createDisabled = $derived(
 		label === '' || amount === 0 || (isRecurring && totalAmount === 0),
 	)
-	const formType = $derived(transaction.id ? 'edit' : 'create')
+	const formType = $derived(transaction?.id ? 'edit' : 'create')
```
vojtechsimetka:
```suggestion
	const formType = $derived(transaction ? 'edit' : 'create')
```

vojtechsimetka (approved):

PR #413, opened 2025-01-27
fix: add client page- added error state

### LINKED ISSUE #401: Any error when creating a client is silently ignored
When I try to create new client and client with the next sequencial PK already exists, the app stays on the page but all the input fields reset.

hlavkat:
resolved #401


src/lib/components/edit-client.svelte
```diff
@@ -39,16 +41,20 @@
 		if (!birthDate) {
 			return
```
vojtechsimetka, line 44: We should also set the error here.

src/lib/components/edit-client.svelte
```diff
@@ -39,16 +41,20 @@
...
+		try {
+			const client: ClientNoId = {
+				name,
+				birth_date: birthDate.toDateString(),
+			}
+			await adapter.addClient(client)
+			name = ''
+			birthDate = undefined
+			imageURI = undefined
+			error = undefined
```
vojtechsimetka: This should be at the start of the function

vojtechsimetka (changes requested):
The error should always be set to `undefined` first.

src/lib/components/edit-client.svelte
vojtechsimetka: @hlavkat please also set `error` here

src/lib/locales/en.json
```diff
@@ -3,6 +3,7 @@
 	"name": "Name",
 	"birthDate": "Birthdate",
 	"birthDateError": "Birthdate cannot be in the future.",
+	"birthDateUndefined": "Set your birth date",
```
vojtechsimetka:
```suggestion
	"birthDateUndefined": "Please set valid birthdate",
```

src/lib/locales/cs.json
```diff
@@ -3,6 +3,7 @@
 	"name": "Jméno",
 	"birthDate": "Datum narození",
 	"birthDateError": "Datum narození nemůže být v budoucnosti.",
+	"birthDateUndefined": "Nastav své datum narození",
```
vojtechsimetka:
```suggestion
	"birthDateUndefined": "Nastav správné datum narození",
```

vojtechsimetka (approved):

PR #421, opened 2025-01-27
fix: number of occurrences in edit-transaction component

### LINKED ISSUE #419: Number of occurrences difference on investment
The number of occurrences (and the calculated amount) are different on the investment card and when editing. See on the screenshots, the card shows 241 while during editing it's only 240.

Image

Image

hlavkat:
resolved #419 

rename function to ```calculateNumOccurrences``` instead of numOccurrences https://github.com/snaha/kalkul/pull/385#discussion_r1924349289


src/lib/components/edit-transaction.svelte
```diff
@@ -51,7 +52,16 @@
 	let endDate = $state(new Date())
 	let period = $state(30)
 	let periodUnit: TimeUnit = $state('year')
-	const numOccurences = $derived(dateDifferenceFunction(repeatUnit)(endDate, date))
+	const numOccurences = $derived(
```
agazso: Please also fix the typo in the variable name:

```suggestion
	const numOccurrences = $derived(
```

src/lib/components/transaction-card.svelte
```diff
@@ -23,7 +23,7 @@
 	let openTransaction = $state(false)
 
 	const numOccurences = $derived(
```
agazso: Please also fix the typo in the variable name:
```suggestion
	const numOccurrences = $derived(
```

vojtechsimetka (changes requested):
This does not fix the issue. Here either the red is incorrect (the date should be 1 month less and it should show 360 occurances and 14,400,000 EUR) or the green is incorrect and should be 361.
Screenshot 2025-01-27 at 15 07 55

hlavkat:
when we change the periodUnit, the end date is recalculated, the problem is when I deposit daily for one month, e.g. from 29.1. so in the end the date will be 28.2. likewise for the deposit date 30.1. etc...

src/lib/components/ui/input/date/picker.svelte
```diff
@@ -22,12 +24,18 @@
...
 
 	let showDatePicker = $state(false)
 
 	const labelVariant = $derived(dimension === 'small' ? 'small' : 'default')
-	const dateString = $derived(value ? value.toISOString().slice(0, 10) : '')
+	const dateString = $derived(value ? formatDate(value) : '')
+
+	$effect(() => {
+		value //eslint-disable-line @typescript-eslint/no-unused-expressions
+		if (onchange) untrack(() => onchange())
```
vojtechsimetka: Maybe this should be a bit more precise because I am unsure if `onchange` will always be untracked. See https://github.com/sveltejs/svelte/issues/9451#issuecomment-1811369546

```suggestion
		untrack(() => { if (onchange) onchange() })
```

src/lib/components/edit-transaction.svelte
```diff
@@ -51,8 +52,15 @@
...
 	let periodUnit: TimeUnit = $state('year')
-	const numOccurences = $derived(dateDifferenceFunction(repeatUnit)(endDate, date))
-	const totalAmount = $derived(numOccurences * Number(amount))
+	const numOccurrences = $derived(
+		calculateNumOccurrences(
+			formatDate(date),
+			formatDate(endDate) ?? formatDate(date),
+			repeatUnit ?? 'month',
+			repeat ?? 1,
+		),
```
vojtechsimetka, line 61: I would prefer that the function takes optional arguments rather than providing defaults here. This can be tackled as a separate PR.

vojtechsimetka (approved):

Issue #423, opened 2025-01-27
Ensure all uses of `window` are behind `if (browser)` check


vojtechsimetka:
Each page may be pre-rendered on the backend where `window` object does not exist. This can cause 500 error code.

For things like location we should use `page` store instead (linked to #469)

PR #427, opened 2025-01-27
fix: app state when user signs out

### LINKED ISSUE #425: Sign out seemingly does not work until refresh
When user signs out, they are redirected to `/` but the page still shows their clients. It only shows the correct page once they refresh. This is likely broken by #369

hlavkat:
resolved #425


vojtechsimetka (changes requested):
This is not what I had in mind. This reloads the page erasing all stores (which is what we want to do but we should erase the stores in signout function). Also it uses the `window` object which is not what we want to do #423

vojtechsimetka (approved):

PR #428, opened 2025-01-27
fix: remeber open investment and transaction

### LINKED ISSUE #405: Remember which investments & transactions are open after adding/editing transaction
Currently all the investments get collapsed

Potentially the solution is as simple as changing the way the investment list is hidden so that they don't re-render (which is what `{#if ...}` statement causes). `display: none` should do the job.

https://github.com/user-attachments/assets/5e425dce-ac66-489c-aee4-080609675b62

hlavkat:
resolved #405


src/lib/components/transaction-card.svelte
```diff
@@ -149,10 +151,16 @@
...
+			display: none;
+		}
 	}
 	.transaction-control {
 		display: flex;
 		padding: var(--half-padding) 0 var(--half-padding) var(--double-padding);
 		gap: var(--half-padding);
+		&.hidden {
+			display: none;
+		}
```
vojtechsimetka: I think you can extract the `hidden` . It does not need to be nested and we can then have it defined just once.

vojtechsimetka (approved):

PR #437, opened 2025-01-29
feat: show hide investment feature

### LINKED ISSUE #376: Implement investment show/hide feature

hlavkat:
resolved #376 

There is an issue on the view page - store transactions are not loading in the investmentCard component


src/lib/stores/investments-view.svelte.ts
```diff
@@ -0,0 +1,66 @@
...
+			hiddenInvestments.set(id, true)
+		}
+	}
+
+	function toggleFocus(id: number) {
+		if (hiddenInvestments.size === allInvestments.length - 1 && !hiddenInvestments.has(id)) {
+			reset()
+		} else {
+			reset()
+			allInvestments.map((inv) => {
```
vojtechsimetka: `map` creates a copy of the array which we don't use at all here
```suggestion
			allInvestments.forEach((inv) => {
```

src/lib/stores/investments-view.svelte.ts
```diff
@@ -0,0 +1,66 @@
...
+		} else {
+			hiddenInvestments.set(id, true)
+		}
+	}
+
+	function toggleFocus(id: number) {
+		if (hiddenInvestments.size === allInvestments.length - 1 && !hiddenInvestments.has(id)) {
+			reset()
+		} else {
+			reset()
```
vojtechsimetka: I don't think you need `reset` here. You can just use `else` in the loop below (more computationally efficient). I would also then ask you to remove the `reset` function because it will be used just once.

src/lib/components/investment-card.svelte
```diff
@@ -34,15 +34,27 @@
 		portfolio: Portfolio
 		viewOnly?: boolean
 		index: number
+		hidden: boolean
+		focused: boolean
 		openTransaction?: (investment: InvestmentWithColorIndex, transaction?: Transaction) => void
+		toggleHide: (e: Event) => void
```
vojtechsimetka: I don't think you need the event here, right?
```suggestion
		toggleHide: () => void
```

PR #440, opened 2025-01-29
feat: add client email

### LINKED ISSUE #433: Add client email
Client email is required

Image

hlavkat:
resolved #433


src/lib/components/edit-client.svelte
```diff
@@ -24,16 +24,32 @@
...
 	let birthDate: Date | undefined = $state()
 	let imageURI: string | undefined = $state()
+	let email = $state('')
 
-	let createDisabled = $derived(name === '' || !birthDate || birthDate > date)
+	let createDisabled = $derived(
+		name === '' ||
+			!birthDate ||
+			birthDate > date ||
+			!/^[\w-]+(\.[\w-]+)*@([\w-]+\.)+[a-zA-Z]{2,7}$/.test(email),
```
agazso: Please put the regex in a const or in a function, so that it can be reused

vojtechsimetka: Actually we should use `zod` here. We do that for other inputs already.

src/lib/components/edit-client.svelte
```diff
@@ -135,6 +161,24 @@
 			invalidDate: $_('Invalid date'),
 		}}
 	></DateInput>
+	{#snippet emailErrorSnippet()}
+		{#if emailError?.email?._errors}
+			{#each emailError?.email?._errors as error}
```
agazso: It would be better to just write `Invalid email address` than to display a bunch of technical errors to the user.

src/lib/components/edit-client.svelte
```diff
@@ -24,16 +26,28 @@
...
+	let emailValid = $state(false)
+	let emailTouched = $state(false)
+	let createDisabled = $derived(name === '' || !birthDate || birthDate > date || !emailValid)
+
+	function onEmailBlur() {
+		if (email.trim() === '') {
+			emailTouched = false
+		} else {
+			emailTouched = true
+		}
```
vojtechsimetka: This is a bit too complicated. Instead I would just do `oninput` and whenever user inputs anything, we change the `emailTouched` to true. Even if user erases everything from the input, it should remain as `emailTouched = true`

src/lib/components/edit-client.svelte
```diff
@@ -89,6 +104,17 @@
 	function confirmDeleteClient() {
 		showConfirmModal = true
 	}
+
+	$effect(() => {
+		const res = emailFormSchema.safeParse({ email })
+		if (res.success) {
+			emailError = undefined
+			emailValid = true
```
vojtechsimetka, line 104: Not sure this is needed but it probably should be here
```suggestion
			emailError = undefined
			emailValid = true
			emailTouched = false
```

vojtechsimetka (approved):

PR #443, opened 2025-01-30
feat: new design of transaction card

### LINKED ISSUE #431: Implement new deposit/withdrawal card style

hlavkat:
resolved #431


src/lib/components/transaction-card.svelte
```diff
@@ -64,58 +74,90 @@
...
+			</Button>
+		{:else}
+			<div class="control-buttons">
+				<Button
+					variant="solid"
+					dimension="small"
+					onclick={(e: Event) => {
+						e.stopPropagation()
+						hidden = !hidden
+						alert('Not implemented in the chart')
```
agazso: Please consider using the `notImplemented` function. It is meant to be used as a placeholder that's easy to search for before a release. It also automatically stops event propagation.

agazso (approved):
Nice!

agazso:
Bonus points for animating the buttons:

You can check this for a potential solution: https://github.com/snaha/kalkul/pull/415
It uses the `animation` property, animating the `opacity`. In our use case besides the opacity the visibility also needs to be changed to make the buttons disappear and not be clickable.

vojtechsimetka (approved):

PR #449, opened 2025-01-31
feat: store feedback in DB

### LINKED ISSUE #314: Feedback button is not implemented
Let's not forget to remove the button in case we don't implement it before releasing

hlavkat:
resolved #314 

#440 must merge first

supabase automatically protects against sql injection
https://github.com/orgs/supabase/discussions/1452


vojtechsimetka:
@hlavkat please resolve conflicts

vojtechsimetka (changes requested):
Looking at this again I think I would love to do this differently. Instead of creating a function that gets client email, it would be better to:

a) create a trigger that on feedback record creation figures out and adds the email (preferred option as it makes it less prone for attack)
b) create a function that creates a feedback record and adds the email

This way we don't need an API endpoint.

src/lib/components/feedback-modal.svelte
```diff
@@ -5,13 +5,35 @@
...
+	let { oncancel, open = $bindable(false), ...restProps }: ModalProps = $props()
 	let message = $state('')
 	const sendDisabled = $derived(message.trim() === '')
+	let error: string | undefined = $state()
+
+	async function sendFeedback() {
+		error = undefined
+		try {
+			const email = authStore.user?.email || null
+			const pathname = browser ? window.location.pathname : null
```
vojtechsimetka: please use the `page.pathname` instead of `window.location`.

src/lib/components/feedback-modal.svelte
```diff
@@ -5,13 +5,35 @@
...
+
+			const { error } = await supabase.from('feedback').insert({ email, message, pathname })
```
vojtechsimetka: Now that this is on a client side, I think we can just add it as a method to `adapter`. This way we don't have to `export supabase` object.

vojtechsimetka (approved):

PR #455, opened 2025-02-04
fix: graph resize

### LINKED ISSUE #355: Graph sizes size vs. window resize
In short they grow when you increase the window size but don't shrink when you reduce. Try it it feels buggy.
Screenshot 2025-01-10 at 15 28 57

hlavkat:
resolved #355


vojtechsimetka (approved):
very nice!

PR #456, opened 2025-02-04
chore: update node to v22 in check workflow

### LINKED ISSUE #302: Update to nodejs22

hlavkat:
resolved #302


vojtechsimetka:
Waiting for merging this until we have implemented the same version requirements in other projects so that we don't have to constantly switch versions.

PR #466, opened 2025-02-17
feat: stripe payments

agazso:
This PR adds payments integration with Stripe. The chosen solution was to use backend API endpoints to communicate with the Stripe API because `stripe-js` (frontend integration) was fragile with the CSP headers and it was also limited what it allowed.

Notes
- The payments functionality in the PR preview is not working, because there is no backend, so all backend calls just return the `index.html`.  The only available functions are the Account functions. In the meanwhile this PR can be tested locally and once it's merged then on the prod website.
- The current solution creates customers (== clients) in Stripe and the subscription management also happens in Stripe. The customers are identified by their emails. There is no subscription related information stored in our database. This means that if the Stripe backend is not working users can not access the app. This could be potentially improved by caching subscription information in our database and if Stripe is not accessible, we could fall back to check our database. This requires (much) more work though and is not in the scope of this PR.
- Because we don't store subscription data in our database we don't need to use the Stripe webhooks, so I removed that (it was in an early version). We may consider having it when we start caching subscription data in our database. 
- Currently we have an early bird promotion. The logic is that it is offered to people who has no subscription yet. However people with canceled trials have at least one subscription, so if they want to buy the product has to buy at full price. I consider this an edge case for now, but we need to have more formal description of promotion and packages to be able handle these kind of edge cases. 
- Also when we remove the early bird option we will need to change the code. In order to avoid that we may want to introduce [Feature flags](https://en.wikipedia.org/wiki/Feature_toggle), but these will make testing more complex.
- The current setup uses a test account in Stripe, so real credit cards will not work, you need to use this card number instead `4242424242424242` (see more [Test payment simulations](https://docs.stripe.com/billing/quickstart#testing)). The other details does not matter, except the card expiry. Expiry and time related functionality can be also tested from the Stripe Dashboard. 
- Currently checking if the subscription is valid happens on the client side, so it's possible to hack it. It should be somehow integrated with the supabase auth, but the [documentation](https://supabase.com/docs/guides/auth/server-side) for that is a mess and somehow they integrated it with SSR. I managed to get the JWT token work with the `Authorization` header, but not with cookies.
- Delete account is not implemented, because there are questions about how it should work. If we delete a customer from Stripe, then all data is deleted and there is no undo for that and that means it may become very cumbersome if someone wants a refund after that. Most likely we should not delete from Stripe at least for a month after the user asks for deleting their data, but the assumes some kind of periodical cleanup process which is outside of the scope of this PR.
- It would be nice to have direct link to the Payments tab under the Account, so that the Stripe checkout can return to it. This can be done in a separate PR.
- We display a text that we send an email 7 days to users before their trial expires. Currently we have no email sending functionality, also it's a question if this should be handled by Stripe or by us. This can be done in a separate PR.
- CSP headers should be added for security reasons. This can be done in a separate PR.


.env
```diff
@@ -1,2 +1,8 @@
 PUBLIC_SUPABASE_URL=http://127.0.0.1:64321
 PUBLIC_SUPABASE_ANON_KEY=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJzdXBhYmFzZS1kZW1vIiwicm9sZSI6ImFub24iLCJleHAiOjE5ODM4MTI5OTZ9.CRXP1A7WOeoJeXxjNni43kdQwgnWNReilDMblYTn_I0
+PUBLIC_ORIGIN='http://localhost:5174'
```
vojtechsimetka: Let's use the default port or we can change the port manually for local testing to e.g. 5200
```suggestion
PUBLIC_ORIGIN='http://localhost:5173'
```

.github/workflows/deploy-prod.yaml
```diff
@@ -7,6 +7,8 @@ on:
 env:
   PUBLIC_SUPABASE_ANON_KEY: ${{ secrets.PUBLIC_SUPABASE_ANON_KEY }}
   PUBLIC_SUPABASE_URL: ${{ secrets.PUBLIC_SUPABASE_URL }}
+  STRIPE_API_KEY: ${{ secrets.STRIPE_API_KEY }}
+  PUBLIC_ORIGIN: ${{ vars.PUBLIC_ORIGIN_PROD }}
```
vojtechsimetka, line 11: I think these need to also be added here 

https://github.com/snaha/kalkul/blob/main/.do/app.yaml#L22-L33

src/lib/components/ui/horizontal.svelte
```diff
@@ -1,14 +1,22 @@
 <script lang="ts">
 	import type { Snippet } from 'svelte'
+	import type { HTMLAttributes } from 'svelte/elements'
 
 	type Props = {
-		children: Snippet
+		children?: Snippet
```
vojtechsimetka: I think we can keep the `children` as required. It does not make sense to have `<Horizontal />`  empty element

vojtechsimetka (commented):
Very nice, thanks for figuring this out! I suggest we quickly fix the two code issues and merge to have this on "production". This way David can also go through this and we can tackle the issues as separate PRs.

As a newly registered customer, I register and the app asks me to confirm email. I go to the email client, click the confirmation link and it correctly opens the app on the billing page. However, in the original tab I bypass the payment page and it loads on the client screen.
Screenshot 2025-03-20 at 8 38 17


When I register as a new customer, this gets logged on the server. I understand why, but I wonder if we need to log this whole error or just indicate that new customer registered and there is no payment attached.
Screenshot 2025-03-20 at 8 34 34

agazso:
> As a newly registered customer, I register and the app asks me to confirm email. I go to the email client, click the confirmation link and it correctly opens the app on the billing page. However, in the original tab I bypass the payment page and it loads on the client screen. 

I fixed all the issues that you found, except this one because I cannot reproduce it. For me the original tab also changes to the payment screen automatically.

PR #471, opened 2025-03-06
build: deploy PR previews as static pages to dev.kalkul.app

### LINKED ISSUE #470: Use static pages for PR previews
This will make PR previews free and after brief testing, the app still works.

We could even consider hosting the whole app statically at least until we build enterprise version (which requires backend for user management).

vojtechsimetka:
Resolves #470

This PR makes PR-previews free because digital ocean does not charge anything for a static pages (up to 3 static pages).

Known issues:
- Refreshing the page on a non-root route in the PR preview causes 404 error (because it uses wrong index.html as fallback). I tried to fix this but there seems to be deliberate limitation in digital ocean.
- Any backend functionality will not work and all backend routes will be stripped.


agazso (approved):
Works well.

The 404 at refresh is annoying, it would be nice if we could work it around.

Also, is using the `base` necessary, or was it an attempt to workaround the 404 issue?

vojtechsimetka:
Yeah the base is necessary because the app is run from a subpath.

The 404 can not be easily solved because digital ocean does not support full ingress rules so I can only point to one index.html as fallback

PR #477, opened 2025-03-07
ci: pr preview use hash router

### LINKED ISSUE #479: Refreshing PR preview gives 404 error
This is because Digital ocean does not allow setting multiple fallbacks (and it has limits on ingress routings).

### LINKED ISSUE #480: Some links use literals instead of constants from routes.ts

vojtechsimetka:
Resolve #479
Resolve #480

We can now easily switch between hash router and normal router. The hash router will only be used for the PR previews


src/lib/routes.ts
```diff
@@ -1,26 +1,32 @@
 import { base } from '$app/paths'
+import { PUBLIC_ROUTER } from '$env/static/public'
```
agazso: I would rather make optional values like this dynamic variable, therefore by default we would not need to provide it at all.

agazso (approved):

PR #489, opened 2025-03-11
fix: buttons in search is optional

### LINKED ISSUE #486: SearchInput improvements
- [ ] search input should have the erase button always visible
- [ ] the erase button should be keyboard navigable

hlavkat:
resolved #486


vojtechsimetka (commented):
It seems that the `SearchInput` component can now be simplified  and all the buttons removed. Please also address #486 at the same time and the position of the search is now on the left.

vojtechsimetka (changes requested):
All good, please just add the `Clear search` button. I would do it outside of the component in the `routes/(app)/+page.svelte` https://www.figma.com/design/CH1QlTIGJk5xblFsNLi3Kn/Kalkul_05?node-id=8301-10789&t=OyIrf1DdxXN6tl5k-1

vojtechsimetka (changes requested):
This is good but please don't show the button if there is nothing in the search input.

vojtechsimetka (approved):

PR #495, opened 2025-03-13
fix: cascade duplicate rollback if fails

### LINKED ISSUE #387: Cascade duplicate should revert if something fails
We don't want to duplicate just part of the data. If any issue happens, we should roll back the duplicated data. Most likely we will need to use some SQL transaction to achieve this. It seems that there is no native support in supabase :/ https://github.com/orgs/supabase/discussions/526

hlavkat:
resolved #387


vojtechsimetka (approved):
good enough for now. Ideally later on these would be done as db transaction but supabase does not support it yet

PR #496, opened 2025-03-13
fix: the portfolio table is modified according figma

hlavkat:
```ROI``` calculation and most likely ```current value``` still need to be completed

### CODE REVIEW

File: src/routes/(app)/client/[id]/+page.svelte [hunk 1]
```diff
@@ -113,6 +116,23 @@
 	</Dropdown>
 {/snippet}
 
+{#snippet viewButton(link: string | null, portfolioId: number)}
+	{#if link}
+		<Button variant="ghost" dimension="compact" href={`${base}/view/${link}`} target="_blank"
+			><Launch size={24} /></Button
+		>
```
Comment (vojtechsimetka) [inline src/routes/(app)/client/[id]/+page.svelte]: If I click this button it correctly opens the view in new window, but it also opens that portfolio in the original tab. Please prevent this.

src/routes/(app)/client/[id]/+page.svelte
```diff
@@ -177,10 +199,14 @@
...
-							<span>{formatDate(new Date(portfolio.last_edited_at))}</span>
-							<span class="right-aligned">{numInvestments(portfolio.id)}</span>
-							<span class="right-aligned">{totalPortfolioDeposits(portfolio.id)}</span>
-							<span class="right-aligned">{totalPortfolioWithdrawals(portfolio.id)}</span>
+							<span>{formatDate(new Date(portfolio.start_date))}</span>
+							<span>{formatDate(new Date(portfolio.end_date))}</span>
+							<span class="right-aligned">{portfolio.inflation_rate * 100}%</span>
+							<span class="right-aligned"
+								>{formatCurrency(portfolioValue(portfolio.id), portfolio.currency)}</span
+							>
```
Comment (vojtechsimetka) [inline src/routes/(app)/client/[id]/+page.svelte:211]: This shows the value as currency first and then value. 
Screenshot 2025-03-13 at 15 49 20

vojtechsimetka (changes requested):

vojtechsimetka (approved):

PR #498, opened 2025-03-17
feat: fee calculation

### LINKED ISSUE #283: Add fees to graph calculation

### LINKED ISSUE #395: Fails to load investment for demo account
I tried to play a bit with Kalkul and created an investment. It looked like the investment was not added but then when I checked the console there is a `DecimalError`. The client has a 20 year long portfolio starting the date they are born. The investment has an APY of `11.53` everything else is 0.

- [ ] We should display some error state if any error happens (and not empty investment page).
- [ ] This should not happen. It seems that this is client side problem and not on DB side.

To test this access (however it may be easier to debug locally):
- https://kalkul.app/client/28/portfolio/20
- `demo@kalkul.app` / `Demo123`

### LINKED ISSUE #530: TER fee on investment
We need [TER (Total Expense Ratio)](https://en.wikipedia.org/wiki/Total_expense_ratio) in addition to current fee structure. The TER should be default setting but users can switch to a more granular fee settings.

Here is my proposal:
Image

Image

### LINKED ISSUE #552: Incorrect calculations
Image
Withdrawal is from March 1998, it is already in the dataset in 1997

### LINKED ISSUE #606: Add investment — Typo on advanced setup
Should be `Advanced setup` not `Advance setup`
Image

vojtechsimetka:
This PR implements the fee calculation for all types of fees that are supported in the UI.

Notes:
- The functionality is extracted into a "library" within the project (`$lib/@snaha/kalkul-maths`) which can be later extracted as a top-level package in the repo.
- It fixes the underlying issue in  #395. One of the issues are that the investment value chart is a line chart so to make it nice it has to have one datapoint longer than the transactions or fees arrays. The fix was just to treat out-of-bound indexes as `0` in the returned arrays.
- The fee storage is unified in database by storing the same value as we display on the UI. Otherwise the UI logic could get quite complicated and error-prone with back-and-forth conversions, especially for values which can be selected as either `fixed` or `percentage`. This is not the nicest, but considering that now we have a library to work with the investments it is pretty much contained.
- The `Frequency` has been renamed to `Period` and a new `PeriodCount` type is introduced when holding a `count` member as well.
- #616
- The fees are aggregated together and displayed as one item, previously the withdrawals were stacked with each having their own inflation values.
- The fees were not added to the breakdown chart, because #617 and it is outside of the scope of this PR. 
- There were some questions during the development, see below the answers as a reference for the future

Questions:
- Is percentage exit fee calculated before the withdrawal or after?
> apply it on top of withdrawal amount. So if user sets to withdraw 100 eur and fee is 5% the total amount will be 105 eur
- Is management fees are calculated before or or after APY?
> After apy (as all fees it is calculated from current portfolio value which first grows due to market growing)
- APY is 0..100, the other percentage values are 0..1
> Leta unify to 0-1

https://discord.com/channels/1217479434270150786/1222654303437848827/1370471110172475593

Fixes #606 
Fixes #395 
Fixes #283 
Fixes #552 
Fixes #530


vojtechsimetka:
Nice, thanks for this @agazso. I found few issues:
- The upfront entry fee does not work correctly (it applies even after the whole fee value was paid)
- Opening the edit investment form always defaults the fee values and types to their default state
- The % exit fee calculation is wrong (for 100k withdrawal it says I paid 2k fee with 0.1% and 208k fee with 10% fee)
- Setting TER with any withdrawal increases the fees horribly
- TER does not seem to apply at all

PR #502, opened 2025-03-19
ci: fix closed pr workflow

vojtechsimetka:


vojtechsimetka:
This is tested and gets rid of the pnpm cache error (because the pnpm install step was skipped but cache was enabled).

PR #505, opened 2025-03-19
fix: remove non-breaking spaces from currency formatting

### LINKED ISSUE #460: Large values can break the app
This is not a huge problem because it is unlikely that these values will be large, but as a general practice it would be good to have a way to solve large values throughout the app. My suggestion would be to use the shorting mechanism with subscript.

This particular one breaks the buttons animation introduced in #452 

Image

hlavkat:
resolved #460


vojtechsimetka (changes requested):
This works for czech locale but not for e.g. the US one (which has numbers in a format `1,000,000` instead of `1 000 000`). I think this is wrong approach TBH and we should instead just do `word-break: break-all` on the relevant components

vojtechsimetka (approved):

PR #508, opened 2025-03-21
fix: trial period remaining days

### LINKED ISSUE #507: trial period remaining days incorrect
Image

hlavkat:
resolved #507


vojtechsimetka (approved):
I think this fixes it, but would love to check with @agazso

agazso:
> I think this fixes it, but would love to check with @agazso

Please check my comment in the original issue: https://github.com/snaha/kalkul/issues/507#issuecomment-2743049700

vojtechsimetka:
@hlavkat please modify this so that when the counter says `0` it will instead be `Expiring today`

vojtechsimetka (changes requested):
Please add text if the number is `0`

agazso (approved):

vojtechsimetka (approved):

PR #528, opened 2025-03-21
feat: disable paywall

### LINKED ISSUE #512: Disable the paywall for the free beta release

agazso:
This PR disables the paywall for the Free Beta release. It disables the paywall and hides the subscribe button in the Account menu for all users except those who have `@example.com` email address.

Fixes #512


vojtechsimetka (approved):
I am OK with this but what if we make it env variable so that we can have environment with and without the payment?

vojtechsimetka (approved):

PR #537, opened 2025-03-25
feat: ```TER``` fee on investment

### LINKED ISSUE #530: TER fee on investment

hlavkat:
resolved #530


vojtechsimetka (changes requested):
This works and thanks for that, but sadly on smaller screens the top of the "add investment" screen is hidden. So are sometimes even the buttons on the bottom. It would be good if it was completely scrollable. Also I think there may be some padding missing at both bottom and top.

https://github.com/user-attachments/assets/e5bfc35a-4e72-45c5-8530-322f503eee3c

src/app.pcss
```diff
@@ -94,6 +94,7 @@ body {
 	background-color: var(--colors-ultra-low);
 	color: var(--colors-ultra-high);
 	font-family: var(--font-family-sans-serif);
+	height: 100%;
```
vojtechsimetka: I think setting it to `vh` is safer

```suggestion
	height: vh;
```

vojtechsimetka: most likely we also want `width: 100vw`

src/lib/components/custom-header.svelte
```diff
@@ -21,8 +21,6 @@
 		display: flex;
 		justify-content: space-between;
 		align-items: center;
```
vojtechsimetka, line 23:
```suggestion
		align-items: center;
		position: sticky;
```

vojtechsimetka (changes requested):
Please add bottom padding 32px

src/app.pcss
hlavkat: When I set body width to 100vw, a problem appeared when there was vertical scrolling on the page, so the width expanded to include the scroll panel and with that, horizontal scrolling also appeared.

vojtechsimetka (approved):

PR #542, opened 2025-03-25
feat: custom tooltip of charts

hlavkat:


src/lib/components/graph-portfolio.svelte
```diff
@@ -153,6 +154,11 @@
 		}
 	}
 	let fullscreenGraph: undefined | 'value' | 'transactions' | 'breakdown' = $state(undefined)
+	let tooltipData: { dataIndex: number; value: number; colorIndex: number; name: string }[] =
+		$state([])
+
+	let tooltipPosition = $state({ x: 0, y: 0 })
+	$inspect(tooltipPosition)
```
vojtechsimetka: Don't forget to remove
```suggestion
```

src/lib/components/graph-portfolio.svelte
```diff
@@ -153,6 +154,11 @@
 		}
 	}
 	let fullscreenGraph: undefined | 'value' | 'transactions' | 'breakdown' = $state(undefined)
+	let tooltipData: { dataIndex: number; value: number; colorIndex: number; name: string }[] =
```
vojtechsimetka: Define this a as a type please

vojtechsimetka (changes requested):
Few issues I found:
- [x] The total still includes hidden investments (it should only sum up the ones that are visible).
- [x] The tooltip position should be right of the cursor for the left half of the graph
- [ ] Maybe we should have `Total (including hidden)`?

agazso:
When moving around the mouse the tooltip is flashing on the default view, ~most likely because between the points there are areas where the tooltip is hidden. My suggestion is to always show the tooltip according to the point which has the smallest distance from the mouse.~

Interestingly there is no flashing on the maximized view where it works exactly how I suggested.

Update: I figured out what causes the flashing. When the mouse is over the tooltip then it no longer triggers the point to display the tooltip, so it hides the tooltip. That triggers the point to display the tooltip again, under the mouse pointer which will hide it again. This is more likely to happen in the small view not maximized view, but I was able to reproduce it on the maximized view as well. I recommend not hiding the tooltip when the mouse is over it, just display it according to the closest point on the graph.

agazso:
Also: do not display the fractional part of the value, only the integer part.

src/lib/utils.ts
```diff
@@ -29,12 +29,12 @@ export function formatAge(birthDate: Date, currentDate = new Date()): string {
 	return age.toFixed()
 }
 
-export function formatCurrency(value: number | bigint, currency: string) {
+export function formatCurrency(value: number | bigint, currency: string, digits: number = 2) {
```
agazso: It would be better to keep the name (e.g. `maximumFractionDigits`) or even better expose an options object. Also we can change the default to be `0`.

vojtechsimetka (approved):

src/lib/components/graph-portfolio.svelte
```diff
@@ -137,6 +139,72 @@
...
+					sumWithInflation += filteredInvestments[j].graphInflationInvestmentValue[i]
+				}
+				totalValue[i] = sum
+				totalWithInflation[i] = sumWithInflation
+			}
+		}
+		return { totalValue, totalWithInflation }
+	}
+
+	function getTotalDeposits() {
```
agazso, line 163: Looks like these calculations could be extracted to reusable functions. Can be done in a different issue/PR.

agazso (approved):
Very nice!

PR #543, opened 2025-03-25
feat: check-locales script

agazso:
This PR adds a script that checks if there are missing translation in a locale and also if there are unused ones.

Usage:
```sh
npx tsx check-locale [locale]
```
By default it checks the Czech locale. It can be also set to `en` (or any other value that has a valid locale file under `src/lib/locales`).

It prints out the keys of the missing translations. It also prints out the unused translations in a file. If there are missing translations or unused translations the script returns with an error.

Once there is proper localisation added we can add a CI check for running the script.


knip.ts
```diff
@@ -6,7 +6,7 @@ const config: KnipConfig = {
 		'$env/*': ['.svelte-kit/ambient.d.ts'],
 		'$lib/*': ['src/lib/*'],
 	},
-	ignore: ['src/lib/typesdb.ts'],
+	ignore: ['src/lib/typesdb.ts', './check-locales.ts'],
```
vojtechsimetka: If you add this as a script in `package.json` it should be found by `knip`.

vojtechsimetka (approved):
Nice! There is a VS code extension that I used for this, but it is good to also have a command line tool!

Issue #586, opened 2025-05-05
Create support@kalkul.app


vojtechsimetka:

vojtechsimetka:
Started spacemail trial and added the login info to protonpass

Issue #584, opened 2025-05-05
New feedback modal (Discord focused)


david-gauquelin:
- Feedback button/popup : what do we want to emphasize? Maybe we can just have an email and Discord

Emphasize using the Discord (we need to clarify with Marek) and maybe have a secondary`support@kalkul.app` email to reach out to

david-gauquelin:
Updated design: https://www.figma.com/design/CH1QlTIGJk5xblFsNLi3Kn/Kalkul_05?node-id=8665-54868&t=puwNjUlZYmvkYTCL-1

<img width="608" alt="Image" src="https://github.com/user-attachments/assets/a642fe1a-e95e-4e5b-8f56-d5c0a52b793a" />

agazso:
Closed as completed

PR #589, opened 2025-05-05
fix: delete account text

### LINKED ISSUE #511: Delete account is not implemented
We can either remove the button or display `Coming soon!`. Implementing at this point would be counterproductive because this will change once the payments are involved. This may be better suited for implementation for the Early Bird release.

agazso:
Fixes #511


src/routes/account/+page.svelte
```diff
@@ -165,7 +165,11 @@
 							>{$_('changePassword')}</Button
 						>
 						<FlexItem />
-						<Button dimension="compact" variant="ghost" onclick={notImplemented}
+						<Button
+							dimension="compact"
+							variant="ghost"
+							onclick={(e: Event) =>
+								notImplemented(e, $_('Contact support to delete your account with all its data'))}
```
vojtechsimetka, line 172: Maybe we should say where. I'd suggest `support@kalkul.app`, wdyt??

vojtechsimetka (approved):

src/routes/account/+page.svelte
agazso: I am waiting for the design of the feedback popup with this (#584) and to have the address setup first (#586). After that we can change the text.

PR #610, opened 2025-05-12
feat: beta landing page

### LINKED ISSUE #598: Coming soon (Get started)
We need a page to prevent people from signing up from the default website

This page should say "coming soon blah blah"
add your email to get notified when it becomes available

vojtechsimetka:
Resolves #598

This only adds the UI part but not the subscribe functionality (issue #614)

TODO:
- [x] consider using `Horizontal` and `Vertical` components
- [x] padding between cta and beta


src/lib/components/beta-landing-page.svelte
```diff
@@ -11,14 +11,33 @@
 	import { notImplemented } from '$lib/not-implemented'
 	import { z } from 'zod'
 	import ErrorComponent from './error.svelte'
+	import Vertical from './ui/vertical.svelte'
 
 	type Props = {
 		isMobile: boolean
 	}
 
+	function sleep(ms: number) {
```
agazso, line 20: This could be in `$lib/utils`

agazso:
I see some differences compared to the design: 
- the bottom part has 32px padding (more precisely, the top part has 16px padding and the whole page `Vertical` has double, see the screenshot)
- the fine print is missing from the design
- the button text is different

Screenshot From 2025-05-13 17-19-35

src/routes/account/+page.svelte
```diff
@@ -62,6 +62,7 @@
 				error = $_('updateLanguageError')
 				console.error('Failed to update language', error)
 			})
+			// FIXME: handle the case when the user is not logged in
```
agazso, line 65: Should there be an issue about this?

agazso (approved):
The code looks good, but see my comments about the differences in the design 

@david-gauquelin

src/lib/components/beta-landing-page.svelte
vojtechsimetka: sure but it is temporary code here

src/routes/account/+page.svelte
vojtechsimetka: Thanks for pointing it out  #618

vojtechsimetka:
I think you have not tried the latest version but some older one. I'll fix the padding and merge.

Screenshot 2025-05-13 at 23 15 38

agazso:
> I think you have not tried the latest version but some older one. I'll fix the padding and merge.

I used the wrong Figma version, sorry.

agazso:
One more thing, the email validity check should not be done on every keystroke, but when you press the button, otherwise it will nag with an error you while you are typing.

vojtechsimetka:
How about running it on blur?

agazso:
> How about running it on blur?

That works too

PR #632, opened 2025-05-14
feat: landing page footer

### LINKED ISSUE #627: Footer on kalkul.app
We still need to understand if we need the disclaimer text or not. If not we keep only the top part of the footer
https://www.figma.com/design/CH1QlTIGJk5xblFsNLi3Kn/Kalkul_05?node-id=8665-18602&t=ePOfVC3kE6hDX0Vq-1

Image
Image
Image

agazso:
Fixes #627


vojtechsimetka:
Looks good but I think the black color should fill the page height, right?
Screenshot 2025-05-14 at 18 38 56

agazso:
> Looks good but I think the black color should fill the page height, right?

Indeed, it should fill the page height and I did not check on a big screen. I made a fix now, but unfortunately it was not a trivial fix, it involved changing the `html` and `body` style sheets so it affects all other pages as well. I tested it somewhat, but it would be good if you could test too as much as possible.

vojtechsimetka (approved):
Tested and it looks good. I'm going through all the pages now so in case I find any problem anywhere I'll create issue

PR #645, opened 2025-05-19
feat: welcome popup

### LINKED ISSUE #628: Welcome pop up
It would be good to welcome users on the desktop app with some message, including some context and a link to the video

https://www.figma.com/design/CH1QlTIGJk5xblFsNLi3Kn/Kalkul_05?node-id=8725-29952&t=ePOfVC3kE6hDX0Vq-1

Image

agazso:
@vojtechsimetka There are new texts, please check the translations.

Fixes #628


vojtechsimetka (approved):
LGTM I'll do the translations in a separate PR.

One thing I noticed is that the video keeps playing even when I close the popup. Sadly it also autoplays more videos after that. Can we do something about it?

agazso:
> One thing I noticed is that the video keeps playing even when I close the popup. Sadly it also autoplays more videos after that. Can we do something about it?

Good catch, I added a fix for this.

vojtechsimetka:
It fixes when I click the button but not when I click outside of the modal.

agazso:
> It fixes when I click the button but not when I click outside of the modal.

I made a new fix that removes the video in all cases when the modal is not open.

vojtechsimetka (approved):

PR #665, opened 2025-05-26
feat: breakdown horizontal chart

### LINKED ISSUE #434: Implement breakdown over time chart
- [x] the slider goes through the chart labels array
- [x] the Doughnut chart is just ratio of investments for a given month/year (index in label array)
- [ ] the Stacked horizontal bar chart is a bit more complex and should be implemented after https://github.com/snaha/kalkul/issues/283

Image

### LINKED ISSUE #616: Fee inflation is not displayed on the transaction chart
The fees are not adjusted with inflation on the transactions chart, because it's not defined how they should be displayed.

My idea is to just extend the fee with the inflation value and use the fee color as well.

### LINKED ISSUE #617: Breakdown chart is incomplete

vojtechsimetka:
resolves #434
resolves #616
resolves #617

This PR improves the performance of the portfolio charts (especially switching between fullscreen and back). It also breaks down the huge `graph-portfolio` file into smaller parts and extracts duplicate code into snippets. It fixes a lot of random known/unknown issues.

TODO:
- [x] add inflation variant to the breakdown chart
- [x] breakdown chart is sometimes wider
- [x] breakdown graph should be initialized to today's date
- [x] custom breakdown tooltips
- [x] translate breakdown tooltips text
- [x] don't show hidden investments in the breakdown chart
- [x] fix tooltip position in the breakdown chart

Not part of this PR:
- How should the fullscreen version of the breakdown chart look like?
- add `interest earned` data to the breakdown chart (should this be before the fees are applied or after?)
- custom tooltip for the doughnut chart (it is broken even in currently deployed version and the custom design is different to other tooltips)
- refactoring of the tooltip mechanism (especially the position calculation which seems wasteful)

Screenshot 2025-05-30 at 0 23 13


agazso (approved):
The code looks good and I found only minor issues that can be addressed as separate PRs:

- In the design the slider is `small`, the implemented version is `compact`
- The label at the end of the slider is not centered vertically, has wrong size and displays the month (which is not needed)

PR #672, opened 2025-05-30
feat: import investment data by isin or ticker

### LINKED ISSUE #176: Actual ISIN data
As an investment advisor I want to download data about an asset from ISIN and not add them manually. I want to search investment either by ISIN number or name

### LINKED ISSUE #678: New ISIN/ticker import modal
This is to implement what we discussed with Marek:

- We should let people import by ISIN or ticker, but we need to provide some help
- We should filter investment by currency to only match the portfolio currency
- We should display the matching products to the user so he can choose from

Also note that the button to call the modal changed design.

https://www.figma.com/design/CH1QlTIGJk5xblFsNLi3Kn/Kalkul_05?node-id=8917-12464&t=zko8tTjiPV0V9FgR-1

agazso:
This PR implements the investment data import by ISIN number or ticker.

Notes:
- The implemented logic is the following: when a user types an identifier, it is sent to the `/api/market/id/{identifier}` endpoint, which checks if it's an ISIN number or a ticker and uses the [OpenFIGI API](https://www.openfigi.com/api/documentation)  to get the name, type, ticker and exchange of the investment. Then this information is returned and the frontend makes another API call to `/api/market/apy/{ticker}` to get the APY. This is not always available because currently we use the [Marketstack API](https://marketstack.com/documentation_v2) which only works with stocks. Adding another API would make this work, but the options found so far was too expensive (see #176).
- The APY is calculated based on the first and last business day of the last year
- There is a script used to generate the data in the `src/lib/exchanges.ts` file. There is some documentation and the there are manual steps in it.

Fixes #678 
Fixes #176


src/lib/locales/en.json
```diff
@@ -160,9 +160,11 @@
 			"addInvestment": "Add investment",
 			"editInvestment": "Edit investment",
 			"investmentName": "Investment name",
-			"importDataInvestment": "Import data for this investment",
+			"importDataInvestment": "Import data for this investmen",
```
vojtechsimetka:
```suggestion
			"importDataInvestment": "Import data for this investment",
```

src/lib/locales/en.json
```diff
@@ -160,9 +160,11 @@
...
 			"importData": "Import data",
-			"isin": "ISIN",
+			"isin": "Please enter an ISIN number or a ticker symbol",
+			"isinInputPlaceholder": "US0378331005 or AAPL",
```
vojtechsimetka, line 166: Feel free to reject this, I just find it clearer
```suggestion
			"isinInputPlaceholder": "e.g. US0378331005 or AAPL",
```

vojtechsimetka (approved):
Great start, thanks! I'd suggest we merge and improve on it. Few things that I encountered:
- for CZK loading ISIN often returns no data but the app is seemingly stuck. It would be good to have some error state)
- it would be good to have an indicator that API request is pending (disabling the button & input)
- missing CZ translations (I'll do it) which breaks the UI a bit now. Maybe this could be fixed as well.
- I am unable to ever load any APY. E.g. for `IE00BYVJRR92` in USD portfolio I get
```
ZodError: [
  {
    "code": "invalid_union",
    "unionErrors": [
      {
        "issues": [
          {
            "code": "invalid_type",
            "expected": "object",
            "received": "undefined",
            "path": [
              "error"
            ],
            "message": "Required"
          }
        ],
        "name": "ZodError"
      },
      {
        "issues": [
          {
            "code": "invalid_type",
            "expected": "number",
            "received": "null",
            "path": [
              "data",
              0,
              "volume"
            ],
            "message": "Expected number, received null"
          },
          {
            "code": "invalid_type",
            "expected": "number",
            "received": "null",
            "path": [
              "data",
              0,
              "adj_volume"
            ],
            "message": "Expected number, received null"
          }
        ],
        "name": "ZodError"
      }
    ],
    "path": [],
    "message": "Invalid input"
  }
]
```
Screenshot 2025-06-11 at 16 05 38

agazso:
>     * for CZK loading ISIN often returns no data but the app is seemingly stuck. It would be good to have some error state)

Opened an issue about it: #688 

>     * it would be good to have an indicator that API request is pending (disabling the button & input)

The right arrow in the button is replaced with a loader and both the button and the input is disabled.

>     * missing CZ translations (I'll do it) which breaks the UI a bit now. Maybe this could be fixed as well.

Opened an issue about it: #689

>     * I am unable to ever load any APY. E.g. for `IE00BYVJRR92` in USD portfolio I get

I fixed this already.

PR #673, opened 2025-06-02
fix: disable analytics on dev

### LINKED ISSUE #661: Disable analytics on dev and preview

agazso:
Fixes #661


vojtechsimetka (commented):
Whz don't we invert this and enable it just on `kalkul.app`? Or even better have the `data-website-id` as public env var and only when provided we enable analytics? This way we can have separate analytics on dev if we need to test something or even on local. And if not enabled, it will not be bundled in the app at all.

agazso:
> Whz don't we invert this and enable it just on `kalkul.app`? Or even better have the `data-website-id` as public env var and only when provided we enable analytics? This way we can have separate analytics on dev if we need to test something or even on local. And if not enabled, it will not be bundled in the app at all.

We can do this, but I chose not to do that for two reasons:
1. It's scope creep. The original issue was about disabling analytics on dev (which the PR does) and not about making the analytics configurable.
2. I find adding environment variables for enabling features requires a lot of busywork with quite a few gotchas: you have to add them for `.env` & `.env.local`, github & digitalocean yamls, then set it on github manually, then set it on digitalocean manually. On digitalocean there are two sets of environments which is confusing and it converts values to yaml values (see #658).

Feel free to do it in a separate PR.

vojtechsimetka:
Ok reasonable. Then my suggestion is to enable it only on `https://kalkul.app`. The idea behind the issue was to disable on dev so that we don't pollute production analytics. This still pollutes it with local host.

agazso:
> Ok reasonable. Then my suggestion is to enable it only on `https://kalkul.app`. The idea behind the issue was to disable on dev so that we don't pollute production analytics. This still pollutes it with local host.

The local dev env is excluded with the first condition (`page.url.protocol === 'http'`) unless you run your local dev env with `https`. I wanted to avoid hardcoding the domain, but I can do that.

vojtechsimetka (approved):

Issue #701, opened 2025-06-14
When realtime DB connection is lost I can create portfolios/investment/transactions but they don't show up and there is no error


vojtechsimetka:

vojtechsimetka:
I can no longer reproduce (nor is it easy to reproduce)
We agreed to close this for now and potentially reopen if the issue ever surfaces https://discord.com/channels/1217479434270150786/1222654303437848827/1394275514902118470

Issue #337, opened 2025-01-09
App is sometimes stuck on loading after editing portfolio/investment info


vojtechsimetka:
This is most likely some DB connection issue. However, the issue often persist through refresh and the only solution is to open the app in another browser. Sometimes opening in new window also works.

agazso:
Anything you can see on the console that can help debugging this issue?

vojtechsimetka:
Sadly not as we were testing it on Marek's computer. But I plan to reproduce this. But it was surprisingly frequent.

Issue #334, opened 2025-01-09
App silently fails to add/edit deposit/withdrawal


vojtechsimetka:
I am unsure how to reproduce this, but surprisingly frequently when I e.g. edit deposit it successfully closes the edit modal but the deposit remains unchanged. Often when this happens, when I try to edit again it again does not save the changes and reverts to original. After few tries or refresh it usually gets fixed. Sadly I've never had console open so I was never able to check it.

I have a suspicion that there is an issue with the database connection.

agazso:
Seems related: #701

vojtechsimetka:
I can no longer reproduce (nor is it easy to reproduce)
We agreed to close this for now and potentially reopen if the issue ever surfaces https://discord.com/channels/1217479434270150786/1222654303437848827/1394275514902118470

PR #724, opened 2025-07-11
fix: remove supabase realtime

agazso:
This PR removes the Supabase Realtime subscriptions, therefore this makes the database operations more simple and reliable.

When deleting or updating it does the operation first to the store, but if it detects an error then it reverts the store to the original state.

This will hopefully improve the following issues:
- #334 
- #337 
- #701


vojtechsimetka (approved):
Very very nice, thanks!

PR #754, opened 2025-07-22
fix: missing translations

agazso:
This PR adds missing translations. It was done with a combination of Claude and `pnpm check-locales`.

It also adds the `check-locales` check to Github workflow. It detects missing translations, unused translations and duplicate entries under one key.


vojtechsimetka (approved):
Wow this is amazing work, thank you!

PR #762, opened 2025-07-30
fix: ISIN import

### LINKED ISSUE #759: ISIN import, showing results in other currencies
We should provide an option to display products in other currencies.

1. When there are matching results in the portfolio currency, we display a text and a button to show all. **Once the user clicks on the button we display all products in a single list and no longer display the text + button. Once the user clicked show all, there's no way to filter them back.**

Image

2. When there are no matching results in the portfolio currency, but we found some matching in other currencies, we display a text and a show products button. I want to keep this step to emphasize that these products are not matching the portfolio settings. **Once the user clicked we display a single list of products, no undo**

Image

3. I added a currency column in the table. This is needed since we will have list displaying results in different currencies

Image

4. I've updated the text content for the following 3 modals, please refer to the Figma:

- https://www.figma.com/design/CH1QlTIGJk5xblFsNLi3Kn/Kalkul_05?node-id=8911-25728&t=uWEyj14eos1jm1qV-1
- https://www.figma.com/design/CH1QlTIGJk5xblFsNLi3Kn/Kalkul_05?node-id=9008-11824&t=uWEyj14eos1jm1qV-1
- https://www.figma.com/design/CH1QlTIGJk5xblFsNLi3Kn/Kalkul_05?node-id=9169-11153&t=uWEyj14eos1jm1qV-1

### LINKED ISSUE #760: ISIN import, no APY data
After importing, if we didn't have data for the APY we should display a message under the APY field.
**It would be great to make that message automatically disappear once people start typing in the input**

https://www.figma.com/design/CH1QlTIGJk5xblFsNLi3Kn/Kalkul_05?node-id=9171-11261&t=uWEyj14eos1jm1qV-1

Image

agazso:
This PR adds several improvements to the ISIN import implementation.

If there is no APY found, then an error message is displayed under the APY field until another value is entered. Fixes #760 

Added results page in case of different currencies. Also fixed a bug that caused not found results in some cases. Fixes #759 

There was an issue with index funds not working, now they are found correctly, and the APY is fetched (in case of S&P 500) from another endpoint, than the stocks. This requires manually adding index funds to a mapping in code in order to work correctly, but at least it's possible to do.

Here are some ISIN numbers that were reported and now work better:
- US78378X1072: this is the S&P 500, now it works with APY
- IE00BG36TC12: this is `X MSCI JAPAN ESG 1C` fund which did not work previously. The APY is still not available (#760), but at least the name and type can be retrieved. It can be tested with using CZK as the currency (#759).


agazso:
One idea I had thinking about this is that we could log/save when someone tries an ISIN where there are no APY or in the case of an index fund we don't know about. Then we could just regularly go through these ISINs and check them manually and potentially fix them or add them in code.

vojtechsimetka (approved):

PR #771, opened 2025-08-06
feat: formatting numbers in inputs

### LINKED ISSUE #709: Number formatting (in inputs)
To make it more easy to read long numbers

vojtechsimetka:
Resolve #709

Added FormattedNumberInput component with format-on-blur behavior (for on input version see #772)
  - Uses app language locale for consistent number display everywhere
  - Smart decimal separator handling - accepts both . and , during input
  - Character filtering blocks invalid characters while preserving copy/paste
  - 6 decimal place limit prevents floating point precision issues

**Key decisions:**
  - Format on blur instead of live formatting to avoid cursor jumping
  - Use app language locale instead of browser locale for consistency (so setting app language also sets locale)
  - Allow flexible decimal input (both separators) but format with correct locale separator
  - Block typing invalid characters but allow paste with automatic filtering


vojtechsimetka:
decided to use the implementation from #772

Issue #664, opened 2025-05-26
Shorten large numbers


vojtechsimetka:
It is quite likely that users will have plans reaching at least `milion` CZK (often 10M+). Such large numbers take quite a lot of space. This is especially annoying on chart labels especially if toggling some values may make the scale jump (from e.g. thousands to millions). We should consider shortening the numbers.

![Image](https://github.com/user-attachments/assets/2f5a9d66-8e35-4f8c-85a4-00c90a689193)

david-gauquelin:
I agree. Here's what I would do, **based on the smallest number on the scale** (we should be consistent and not mix different styles on the same graph. If the smallest number on the scale (except 0) is:

- under 1,000 then: we display the full numbers
- between 1,000 and 999,999 we use `k` (1k, 1.5k, 50k, 999k, etc.)
-  between 1,000,000 and 999,999,999 we use `m` (1m, 1.5m, 50m, 999m, etc.)
- from 1,000,000,000 and over we use `b` (1b, 1.5b, 50b, 999b, etc.)

vojtechsimetka:
Linked to #391

Issue #404, opened 2025-01-24
Display of long transaction names/amounts in the investment card


vojtechsimetka:
I believe you will not like this :)
<img width="381" alt="Image" src="https://github.com/user-attachments/assets/f4591504-ff39-485b-a50e-8908ecc9f377" />

david-gauquelin:
We should try to not wrap the number and give it the maximum width possible. We should rather wrap the label on multiple lines. If the label reached a minimum size (like 48px for example) and the number is still too long to fit, only then I would wrap the number on 2 lines. But that should be an edge case, especially because I've made the label much smaller in the new design, see below

<img width="609" alt="Image" src="https://github.com/user-attachments/assets/79f5abf1-7708-4d8a-97c5-8f77a64c6a4b" />

See new design for the transactions list in Figma (scroll up/down to see hover states that reveal the actions)
https://www.figma.com/design/CH1QlTIGJk5xblFsNLi3Kn/Kalkul_05?node-id=4165-5409&t=BjCR6QG2uFG0vmtU-1

david-gauquelin:
See https://github.com/snaha/kalkul/issues/429

PR #772, opened 2025-08-06
feat: formatting in inputs as users type

### LINKED ISSUE #684: Strange labels on graphs
How to view:

- log in with the `demo@kalkul.app` account
- select `Vanguardia Steadyworth`
- select `Portfolio 2`
- look at the breakdown graph horizontal labels

Image

### LINKED ISSUE #709: Number formatting (in inputs)

### LINKED ISSUE #717: Inputs should use a 0 placeholder not entered value
I noticed that in multiple places when we have a number input we show it by default with a 0 value that is entered (not a placeholder text) the issue that when I focus on the input the 0 value remains and I have to erase it. I think we should use a placeholder test instead, so that when I start typing (or even just focus on the input) the input field should be automatically cleared.

Image

vojtechsimetka:
Issues fixed:
- resolve https://github.com/snaha/kalkul/issues/709
- resolve #684 (which is somewhat related to #664 and #404)
- resolve #717 (it selects the content if the value is `0` when clicking in the input)

**Features:**
  - Real-time thousand separator formatting during typing
  - Smart cursor positioning when separators are added/removed
  - Intuitive deletion behavior for thousand separators
  - Dual decimal separator support (accepts both . and ,)
  - Character filtering with paste support
  - 15-digit precision limit to prevent floating point issues

**Technical highlights:**
  - Comprehensive Playwright component test coverage
  - Cursor logic based on unformatted numbers for reliability
  - Live formatting instead of format-on-blur for better UX (see #771)


agazso (commented):
I like this approach better than checking on blur.

However I found issues with this PR:
- It does not let me enter a trailing zero, so I cannot type `1.05` because after typing `1.` and then zero it deletes the decimal point.
- When you delete everything the value defaults to `0` and the cursor is put before the `0`, so when you type `1` the value becomes `10`. It would be better to delete everything and implicitly convert the empty string to `0` in the bound value.
- Also when typing `-` it defaults to `0`. If we don't support negative values it would be better to just ignore the `-`.
- If there is a default `0` value the cursor is after the `0`, and I type `1` then the value becomes `1`. This may be a side effect of the empty value being equal to `0`, but it is sure surprising.
- The MAX_DECIMAL_PLACES should be a prop and should have a higher default value, because with this the maximum transaction value is `9,999,999` (in the selected currency) and we should support higher values than that.
-

agazso (commented):
It is better but there are still some issues left:
- Currently it allows maximum 4 digits numbers, so the maximum available transaction value is `9,999`
- The zero handling is better, but it allows to input unlimited zeroes

agazso:
I still have the issues from above (https://github.com/snaha/kalkul/pull/772#pullrequestreview-3097472898)

I also found some new issues:
- An addition to the 4 digits maximum: I can enter longer value if I put the cursor before the first separator, but not in any other positions.
- Selection works inconsistently. For example: enter `1234`, this will become `1,234`. Then select the `23`  and try to enter a new number, nothing happens. If I press backspace then the result will be `234` (delete works fine).
-

src/lib/components/graph-portfolio-breakdown.svelte
```diff
@@ -119,6 +119,9 @@
 		$_('common.totalFees'),
 	]
 	let cursorX = $state(0)
+
+	// Store locale for use in callbacks
+	const currentLocale = $derived($locale)
```
agazso: Is this really necessary? Can't we use `$locale` directly?

src/lib/utils.ts
```diff
@@ -32,24 +32,52 @@ export function formatAge(birthDate: Date, currentDate = new Date()): string {
 export function formatCurrency(
 	value: number | bigint,
 	currency: string,
+	locale?: string | null,
```
agazso: I would consider making this non-optional, because if it's optional we can easily forget it and it looks like there is no such use case when the locale is not provided.

Or alternatively import it from `svelte-i18n` and not have this as argument.

agazso (approved):
Works well!

src/lib/utils.ts
vojtechsimetka: Yes I'll make it non-optional. I originally used the `svelte-i18n` directly but it was hard to get it setup with the playwright tests.

PR #774, opened 2025-08-08
feat: playwright basic setup

vojtechsimetka:


agazso (approved):
It works locally as well on linux.

PR #775, opened 2025-08-08
fix: isin errors migration

agazso:
This PR fixes the migration file and also makes the app ignoring errors that may happen during reporting ISIN errors.


vojtechsimetka (approved):
Nice! Tested locally on previous DB and also fresh DB. Works fine.

PR #777, opened 2025-08-08
feat: isin cache

agazso:
This PR adds caching to the OpenFIGI market data. 

**Requires a manual database migration on production**

Notes:
- There is an expiration of the cached data (currently a month) so if the OpenFIGI data changes (usually markets can be added or removed) then the cached data will be updates. Currently the expiration is one month, this may be changed in the future.
- There is a new adapter for the backend to access the database with secret key instead of an authenticated user. This is using a private environment, so it cannot imported at the frontend.
- I also fixed the error reporting mechanism, because it used an unimplemented endpoint (`/api/market/error/[identifier]`)


.github/workflows/deploy-dev-preview.yaml
```diff
@@ -21,6 +21,7 @@ env:
   PUBLIC_DISCORD_LINK: ${{ vars.PUBLIC_DISCORD_LINK }}
   RESEND_API_KEY: ${{ secrets.RESEND_API_KEY }}
   RESEND_AUDIENCE_ID: ${{ secrets.RESEND_AUDIENCE_ID }}
+  SUPABASE_SECRET_KEY: ${{ secretes.SUPABASE_SECRET_KEY }}
```
vojtechsimetka: Typo

.github/workflows/deploy-prod.yaml
```diff
@@ -14,6 +14,7 @@ env:
   RESEND_API_KEY: ${{ secrets.RESEND_API_KEY }}
   RESEND_AUDIENCE_ID: ${{ secrets.RESEND_AUDIENCE_ID }}
   MARKETSTACK_API_KEY: ${{ secrets.MARKETSTACK_API_KEY }}
+  SUPABASE_SECRET_KEY: ${{ secretes.SUPABASE_SECRET_KEY }}
```
vojtechsimetka: Typo

src/routes/api/market/id/[identifier]/+server.ts
```diff
@@ -29,7 +58,9 @@ export async function GET({ params }) {
 		})
 
 		const jsonResponse = await response.json()
-		console.debug({ jsonResponse })
+
+		// Store in cache
+		await safeAddMarketData(identifier, idType, jsonResponse)
```
Comment (vojtechsimetka) [inline src/routes/api/market/id/[identifier]/+server.ts:63]: Maybe we don't need to have await here? That will make the request a bit faster and this can finish on server asynchronously

supabase/migrations/20250808155338_market_data_cache.sql
```diff
@@ -0,0 +1,29 @@
+create table "public"."market_data_cache" (
+    id bigint generated by default as identity not null,
+    created_at timestamp with time zone not null default now(),
+    updated_at timestamp with time zone not null default now(),
+    identifier text not null,
+    id_type text not null,
+    response_data json not null,
```
vojtechsimetka, line 7: Any chance we could save this as structured data?

vojtechsimetka (approved):

src/lib/adapters/supabase/index.ts
```diff
@@ -378,18 +384,62 @@ export default class Supabase implements Adapter {
...
+		identifier: string,
+		idType: string,
+		updatedAfter: Date | undefined = undefined,
+	): Promise<object | undefined> {
+		const baseQuery = this.supabase
+			.from('market_data_cache')
+			.select('response_data')
+			.eq('identifier', identifier)
+			.eq('id_type', idType)
+		const query = updatedAfter ? baseQuery.gt('updated_at', formatDate(updatedAfter)) : baseQuery
```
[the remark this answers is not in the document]
vojtechsimetka: actually shouldn't this just use integer comparison between two integer timestamps?

supabase/migrations/20250808155338_market_data_cache.sql
agazso: The `json` type in Postgres can be queried as structured data. I think that's enough for now, later we can always add more columns and/or tables and update those from querying the json.

```sql
  SELECT
      identifier,
      elem->>'name' as company_name,
      elem->>'ticker' as ticker,
      elem->>'exchCode' as exchange
  FROM market_data_cache,
  json_array_elements(response_data->0->'data') as elem
  WHERE elem->>'exchCode' = 'US';
```

src/routes/api/market/id/[identifier]/+server.ts
Comment (agazso) [reply src/routes/api/market/id/[identifier]/+server.ts:63]: I thought about this, but I also have traumatic memories of debugging memory leaks :) 

Maybe I am too cautious, but I would not optimize until it does not feel slow. My assumption is that the FIGI call would be at least an order of magnitude slower than writing our database but I haven't tested.

I opened an issue: #781

src/lib/adapters/supabase/index.ts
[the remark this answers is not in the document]
agazso: The Supabase API expects a date string for date types instead of a timestamp.

src/routes/api/market/id/[identifier]/+server.ts
```diff
@@ -18,6 +38,15 @@ export async function GET({ params }) {
 		}
 
 		const idType = identifierIdType(identifier)
+
+		// Only return cached data that is newer than a month
+		const updatedAfter = subMonths(new Date(), 1)
+		const cachedData = await safeGetMarketData(identifier, idType, updatedAfter)
+		if (cachedData) {
+			return json(cachedData)
```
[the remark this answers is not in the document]
Comment (agazso) [reply src/routes/api/market/id/[identifier]/+server.ts:46]: The cache is not critical for the functionality to work and it is logged when there is an error.

PR #778, opened 2025-08-10
test: add playwright component tests for checkbox

vojtechsimetka:
The goal is to have a browser based component testing for building new components like the formatted input component #772 . To get this working, I chose the checkbox as a component that has some interaction but is not as complex as inputs/buttons.

Please note that this uses an alpha pre-release version of the playwright CT library because the svelte 5 support was only merged 5 days ago https://github.com/microsoft/playwright/pull/36918 .


agazso (commented):
The code looks okay, but the tests don't run. Maybe it's because of the alpha version playwright?

agazso (approved):

PR #782, opened 2025-08-11
wip: database security

agazso:
This PR is meant to stay work-in-progress until closed and never merged to production.

The migration contains the necessary SQL statements to generate the database schema, and it works correctly after the database is reset. However in production we need to do a manual, multi-phase migration for this one:

1. Add `user_id` columns to the user data tables (portfolio, investment, transaction). This can be one PR/migration. At this point the `user_id` column can be null or we can store empty strings by default.
2. Create and run a query which populates the `user_id` columns taken from the `client` table transitively. This can be one or several queries, run manually (since this does not modify the schema).
3. Add the RLS policies which rely on the `user_id` to be set correctly. This can be another PR/migration.


agazso:
Closed in favor of #785

PR #788, opened 2025-08-13
fix: show correct todays investment value in client overview table

### LINKED ISSUE #663: Current portfolio value is wrong in client overview
Image

Image

vojtechsimetka:
**Summary**
- Fixed incorrect portfolio value calculation in /client page that was showing deposits minus withdrawals instead of actual current value with compound growth (resolves #663)
- Added `getCurrentInvestmentValue` and `getCurrentPortfolioValue` functions to calculate today's value with proper APY and fee calculations
- Updated portfolio table text to display "Today's value" instead of "Current value" in both English and Czech translations
-  Extracted shared calculation logic into `calculateDailyInvestmentUpdate` to maintain consistency between functions

**Performance**
The `getCurrentPortfolioValue` function creates an object on every call which adds overhead compared to inline calculation. Roughly the benchmark with 10 investments in a portfolio are following:
- 5-year calculation: ~35ms (vs ~13ms inline)
- 20-year calculation: ~150ms (vs ~58ms inline)
- 80-year calculation: ~600ms (vs ~230ms inline)

I don't think we need to be too concerned here and having the logic in one place is more important IMHO . However, I documented this in code so if this becomes a problem we know where it comes from.


agazso (approved):
The code looks good and it works well.

I would not display the fractional part of the total investment value (or anywhere in investment value) at the moment. In the currencies we support at the moment it is meaningless and just makes it harder to read the value.

Maybe when supporting cryptocurrencies this will make more sense, but for that a [smart formatting approach](https://github.com/snaha/ghost.fun/blob/4696401fab388ad6be8522fcf9d608e3cc184dd4/app/src/lib/utils/format.ts#L109) would work better as well.

PR #790, opened 2025-08-13
feat: pdf export

### LINKED ISSUE #465: PDF export
It would be relatively easy to do PDF export by removing unnecessary UI elements and opening the system print dialog. Every major operating system has PDF export option when printing, so it is possible to rely on that.

https://developer.mozilla.org/en-US/docs/Web/API/Window/print

agazso:
This PR implements the PDF export feature.

Only the value chart and the breakdown chart is implemented as a start.

Notes:
- There are minor differences from the designed version. For example the key date selection does not have a default key data, instead pressing the add button adds predefined dates automatically. The proposed alternative in the design is to use the end date of the portfolio, but then it's cumbersome to add an earlier date, unless the order of the dates can be easily changed. This will require some iterations based on the feedback to settle on something that's easy to use and not difficult to implement either.
- There is also a difference from the design in size. The figma assumes that the A4 page `842x595` whereas in the browser it's `1122x793`, hence there is a new CSS constant added called `--print-multiplier` to compensate. Ultimately the best would be to use print units or `mm`s.
- The flexible breakdown columns are implemented for 1-4 columns, however the edge-case when the number of investments is too long to fit on one page is not implemented as the design. 
- There is a new localized date input component added and existing instances of date input were replaced by it.
- The PR was mostly done by Claude (besides tweaking the sizes at the end) because I wanted to improve it to work with Diete. The updates are now saved to `CLAUDE.md` and with that Claude is pretty capable of using Diete correctly.

Fixes #465


src/lib/components/pdf-export-header.svelte
```diff
@@ -0,0 +1,67 @@
...
+
+	interface Props {
+		portfolioName: string
+		viewLink?: string
+	}
+
+	let { portfolioName, viewLink }: Props = $props()
+
+	// Get current locale for date formatting
+	const currentLocale = $derived(get(locale) || 'cs')
```
vojtechsimetka: We can just use $locale directly, right?

src/lib/components/pdf-export.svelte
```diff
@@ -0,0 +1,397 @@
...
+	let {
+		graphValueData,
+		adjustWithInflation,
+		clientBirthDate,
+		portfolioName,
+		keyDates = [],
+	}: Props = $props()
+
+	// Get current locale for date formatting
+	const currentLocale = $derived(get(locale) || 'cs')
```
vojtechsimetka: Probably can use $locale directly (lets mention this in claude.md, this is recurring pattern).

src/lib/components/pdf-export.svelte
```diff
@@ -0,0 +1,397 @@
...
+
+	// Get portfolio and construct view link
+	const portfolio = $derived(graphValueData?.portfolio)
```
vojtechsimetka: Does this need to be derived?

vojtechsimetka (commented):
I can't do a proper review now, but i wonder whether some things in the ui components should not be passed as props. This would help decoupling it and maybe we can avoid some derived stuff.

src/lib/components/pdf-export.svelte
agazso: I fixed the locale and updated claude.md

src/lib/components/edit-client.svelte
```diff
@@ -144,25 +144,16 @@
...
 			error={birthDate && birthDate > date ? birthDateError : undefined}
-			errorMessages={{
-				invalidYear: $_('error.invalidYear'),
-				invalidMonth: $_('error.invalidMonth'),
-				invalidDay: $_('error.invalidDay'),
-				invalidDate: $_('error.invalidDate'),
-			}}
-		></DateInput>
+		/>
 		<Input
```
vojtechsimetka, line 157: Are these changes here on purpose? Seems unrelated to the PR.

src/lib/components/pdf-export-footer.svelte
```diff
@@ -0,0 +1,46 @@
...
+
+	let { pageNumber, totalPages }: Props = $props()
+</script>
+
+<footer class="chart-footer">
+	<Horizontal --horizontal-justify-content="space-between">
+		<Horizontal --horizontal-gap="0">
+			<Logo size="32" color="var(--colors-neutral)" />
+			<Typography style="font-size: 10.66px; ; line-height: 13.325px;">
+				Made with <a
```
vojtechsimetka: Missing translation. It would be "Vytvořeno v **Kalkul.app**"

vojtechsimetka (approved):
I am fine merging this as is and building on top of this. I think it's missing the transaction graph, right?

This works well on Chrome, but not so much on Safari - the page orientation is by default portrait so the graph etc is just in the top of each page. Switching to landscape mode adds empty pages

src/lib/components/edit-client.svelte
agazso: Yes, the PDF export uses the `DateInput`, which needs to be localized, so I created a localized version of it. See the notes.

agazso:
> I am fine merging this as is and building on top of this. I think it's missing the transaction graph, right?
> 
> This works well on Chrome, but not so much on Safari - the page orientation is by default portrait so the graph etc is just in the top of each page. Switching to landscape mode adds empty pages

Yeah, it's missing the transaction graph, because Marek said it's less important. My approach is to first complete and polish the important things, then later we can add less important features.

I opened an issue about the Safari portrait mode: #806

PR #799, opened 2025-08-14
fix: authorizedFetch refresh

### LINKED ISSUE #798: authorizedFetch does not refresh session
It may happen that you leave for hours, so that the session expires and later you try to access a backend functionality with `authorizedFetch`. For example you leave the import investment dialog open for an hour and try to fetch a stock, it will fail silently.

The `authorizedFetch` should be able to refresh the session.

agazso:
Fixes #798


vojtechsimetka (approved):
Is there some way to test this? Either with vitest or playwright ideally.

agazso:
> Is there some way to test this? Either with vitest or playwright ideally.

Not easy without mocking the whole supabase auth, otherwise it will need to wait the session timeout.

PR #800, opened 2025-08-14
test: synthetic seed data

agazso:
Not to be merged, only for testing purposes!


vojtechsimetka:
TBH I think we could merge this. The current seed data is horrible and should be improved greatly.

PR #805, opened 2025-08-15
feat: animate tooltip

agazso:
This PR adds a small animation to the tooltip.


src/lib/components/tooltip-base.svelte
```diff
@@ -41,11 +41,12 @@
...
-	<div
-		class="tooltip"
-		style={`transform: translate(calc(${tooltipPosition.x}px - 100%),calc(${tooltipPosition.y}px - 50%)`}
-	>
+<div
+	class="tooltip"
+	class:visible={tooltipData.length > 0}
+	style={`transform: translate(calc(${tooltipPosition.x}px - 100%),calc(${tooltipPosition.y}px - 50%)`}
+>
+	{#if tooltipData.length > 0}
```
vojtechsimetka, line 49: Do we still need this if?

vojtechsimetka (approved):

src/lib/components/tooltip-base.svelte
agazso: Yes, this is is needed, otherwise the children rendering will be broken. The idea was to have an outside wrapper which is not unmounted and mounted every time a data point is hovered, but the wrapper can stay and only the visibility changes.

vojtechsimetka: ok, feel free to merge :)

Issue #807, opened 2025-08-15
Formatted input validation should not overwrite the incorrect value


vojtechsimetka:
Instead it should show the wrong value and some error message. This is how other libraries do it.

PR #808, opened 2025-08-15
[wip] fix: bound value in formatted input

### LINKED ISSUE #792: When editing a transaction the amount is empty
It should show the amount of the transaction

vojtechsimetka:
Resolves #792

However, this PR breaks the input validation mechanism from the bound value itself. While this is not a big issue (it is unlikely that we will populate the bound value with an incorrect value), our constraint handling is different than default html handling or other libraries handling. I created an issue about it #807


vojtechsimetka:
implemented as part of #809

PR #818, opened 2025-08-18
fix: investment card menu dropdown

### LINKED ISSUE #563: Menu button on investment card has wrong size/proportions
Image

Also I'm not sure that I like the showing only on hover, I would prefer to remove this and show the button all the time.

agazso:
Also fixes dropdown default behavior.

Fixes #563


src/lib/components/investment-card.svelte
```diff
@@ -207,9 +207,9 @@
 		}
 	}
 	.dropdown {
-		opacity: 0;
-		transition: opacity 0.2s ease-in;
-		visibility: hidden;
+		opacity: 1;
+		transition: opacity 0.2s ease-out;
+		visibility: visible;
```
vojtechsimetka, line 212: do we need this at all? Seems unnecessary since there is no opacity change ever and it is always visible

vojtechsimetka (approved):

PR #814, opened 2025-08-18
fix: edit transaction sidebar scrolling issues

agazso:
Fixes #763

PR #819, opened 2025-08-18
test: add delay to all pressSequentilly to ensure tests are not flaky

vojtechsimetka:


agazso (approved):
Do you suspect this causes the issue with #814?

vojtechsimetka:
@agazso yes that is what I think

PR #834, opened 2025-08-21
fix: client email improvements

### LINKED ISSUE #604: Add client — Improvements on email field
1. Change the placeholder text to `Client's email` instead of `Email`
2. Check on the email format only when pressing the `Create client` button (currently it's checking as I type which pretty annoying)
3. I would consider to not make it mandatory see #603

agazso:
Fixes #604


src/lib/locales/en.json
```diff
@@ -335,7 +335,8 @@
 		"month": "Month",
 		"day": "Day",
 		"email": "Email",
-		"emailOptional": "Email (Optional)",
+		"emailOptional": "Client's email (Optional)",
+		"clientEmail": "Client's email",
```
vojtechsimetka: This seems to be missing in CZ translation?

vojtechsimetka (approved):

src/lib/locales/en.json
agazso: This was not used at all. I added an extra check to the `check-locales` script to catch these cases.

PR #838, opened 2025-08-22
feat: add sorting to columns on the homepage

agazso:


src/routes/(app)/+page.svelte
```diff
@@ -229,11 +295,31 @@
...
+					<!-- svelte-ignore a11y_click_events_have_key_events -->
+					<!-- svelte-ignore a11y_no_static_element_interactions -->
+					<span class="right-aligned sortable" onclick={() => setSortColumn('portfolios')}
+						>{@render sortIcon('portfolios')} {$_('common.portfolios')}</span
+					>
+					<!-- svelte-ignore a11y_click_events_have_key_events -->
+					<!-- svelte-ignore a11y_no_static_element_interactions -->
+					<span class="right-aligned sortable" onclick={() => setSortColumn('investments')}
+						>{@render sortIcon('investments')} {$_('common.investments')}</span
+					>
```
vojtechsimetka, line 322: These should be buttons.

vojtechsimetka (approved):
Nice! I have one design feedback - it is not obvious that the columns are sortable. Usually there are is a up & down caret on all columns.

agazso:
For now we just order by creation date: #890

PR #839, opened 2025-08-24
feat: inflation adjusted transactions

### LINKED ISSUE #671: Refactor the tooltip mechanism
Especially the position calculation seems to be wasteful

### LINKED ISSUE #710: Automatically adjust transactions with inflation
It would be good to be able to automatically adjust the withdrawals by inflation. For example when planning retirement (e.g. long term monthly withdrawals)

vojtechsimetka:
You can test this with a demo account which has inflation adjusted portfolio in the default client.

### Open questions
- Should the inflation adjustment be on by default?
- How should we improve the visuals? (and can we merge as is and improve later?)

### Resolves #710
- Adds inflation adjusted transaction with inflation applied daily (just as all the other transactions we have)
- The edit transaction now has a toggle that switches the transaction valuation on (see screenshots below)
- The edit transaction screen now always shows the nominal total value (see screenshots below)
- The transaction card now shows also nominal values if show inflation is on (see screenshots below)
- There is an indication in the transaction card to show that a given transaction has inflation adjustment on (see screenshots below)

### Resolves #671 (I could extract this into separate PR if needed)
- The tooltips positioning is now done with CSS (the JS implementation before was wasteful and did not work for inline components)
- The tooltip content now wraps when it reaches max width of 500 pxs
- There is now a detection whether the tooltip fits on the screen and if not it chooses different position with following logic:
    - For top position: Try top → bottom → left → right → bottom (with overflow)
    - For bottom position: Try bottom → top → left → right → top (with overflow)
    - For left position: Try left → right → top → bottom
    - For right position: Try right → left → top → bottom

### Other improvements:
- use everywhere `formatDate` from `$lib/@snaha/kalkul-maths/date` for consistency
- 

Edit transaction single total value (show inflation is off)
Screenshot 2025-09-03 at 18 50 26

Edit transaction with both nominal and adjusted value  (show inflation is on)
Screenshot 2025-09-03 at 18 50 35

Transaction card with single total value (show inflation is off)
Screenshot 2025-09-03 at 18 50 56

Transaction card with both nominal and adjusted value  (show inflation is on)
Screenshot 2025-09-03 at 18 50 49

Inflation adjusted transaction tooltip
Screenshot 2025-09-03 at 18 50 09


src/lib/locales/en.json
```diff
@@ -296,8 +295,8 @@
 			"editTransaction": "Edit transaction",
 			"duplicateTransaction": "Duplicate transaction",
 			"deleteTransaction": "Delete transaction",
-			"totalDeposit": "Total deposits: {amount} {currency}",
-			"totalWithdrawal": "Total withdrawals: {amount} {currency}"
+			"totalDeposit": "Total deposits: {amount}",
+			"totalWithdrawal": "Total withdrawals: {amount}"
```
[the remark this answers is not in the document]
vojtechsimetka: Not an issue (this is consistent with other translations and actually fixes an error)

src/lib/@snaha/kalkul-maths/calculation-utils.test.ts
```diff
@@ -21,3 +21,50 @@ describe('#calculateTotalDepositAmount', () => {
...
+
+describe('#calculateInflationAdjustedAmount', () => {
+	it('returns original amount when transaction and base dates match', () => {
+		const result = calculateInflationAdjustedAmount(1000, '2024-01-01', '2024-01-01', 0.03)
+		expect(result).toBe(1000)
+	})
+
+	it('applies positive inflation over one year', () => {
+		const result = calculateInflationAdjustedAmount(1000, '2025-01-01', '2024-01-01', 0.03)
+		expect(result).toBeCloseTo(1030.06, 2)
```
agazso, line 38: Why is this not exactly `1030`? Because of the leap year?

In that case I would suggest to either copy these tests for a non-leap year, or just use a non-leap year.

vojtechsimetka: Even for non-leap years this will be slighly off (`1029.98`) because we use `365.25` days in year (average over leap/nonleap years). This means we don't get exact values but we are statistically very close.

Alternative is to fix this fully and check if a year is a leap year or not (and use `365` and `366`). But that is a larger refactor and will introduce calculation overhead.

vojtechsimetka: In the mean time, I duplicated the tests to run for both leap and non leap years.

src/lib/@snaha/kalkul-maths/calculation-utils.ts
```diff
@@ -7,3 +12,106 @@ export function calculateTotalDepositAmount(deposits: TransactionMap): number {
...
+		const deflationMultiplier = new Decimal(1 + inflationRate).pow(-yearsDifference)
+		const realAmount = new Decimal(calculatedAmount).mul(deflationMultiplier)
+		return realAmount.toNumber()
+	} else {
+		// Transaction is NOT inflation-adjusted: deflate future amounts to present value
+		// When showing inflation, we want to see what future amounts are worth in today's terms
+		const deflationMultiplier = new Decimal(1 + inflationRate).pow(-yearsDifference)
+		const presentValue = new Decimal(calculatedAmount).mul(deflationMultiplier)
+		return presentValue.toNumber()
+	}
```
agazso: The calculation seems to be identical on both branches of the condition, just with different names,

src/lib/@snaha/kalkul-maths/calculation-utils.test.ts
agazso: Well, in the app we don't need this precision anyway, because the inflation is also just an average, however it would make tests more comprehensible.

src/lib/@snaha/kalkul-maths/calculation-utils.ts
vojtechsimetka: Thanks for spotting this! Fixed

src/lib/@snaha/kalkul-maths/calculation-utils.ts
```diff
@@ -7,3 +12,106 @@ export function calculateTotalDepositAmount(deposits: TransactionMap): number {
...
+ * @param showInflation Whether inflation visualization is enabled in UI
+ * @returns The amount to display in the UI
+ */
+export function calculateDisplayAmount(
+	calculatedAmount: number,
+	transactionDate: string,
+	baseDate: string,
+	inflationRate: number,
+	transactionInflationAdjusted: boolean | undefined,
+	showInflation: boolean,
```
agazso: Wherever this function is called it is called twice, once with `showInflation` `true` and once more with `false`. It looks like the `showInflation` can be just omitted and use the `calculatedAmount` when it's false.

Alternatively this function can return both values.

agazso (commented):
A major issue that I found is that the calculations are significantly slower, even though I don't have any inflation adjusted transactions.

With the `Sensible investing` portfolio (which does not have a lot of transactions) the time goes from sub-second to around 2-3 seconds when I press the `Show inflation` switch. This applies also when changing a transaction value or when opening a portfolio. With a more realistic portfolio (around 20-30 transactions) the time can go up to 4-5 seconds.

I think before merging we should at least identify what causes this and have a plan to address it, otherwise the user experience may suffer.

vojtechsimetka:
@agazso you are right that the performance is not great (I don't experience this in my computer but it is surely on the upper bound performance wise). I gave claude some free reign with optimizing the new calculations and IMHO it improved things a lot. LMK what is your experience #874 .

I think we should generally optimize the calculations (because we unnecessarily recalculate things especially whole investments where nothing changed). But I think that should be a separate task which also adds performance metrics. For that it would be good to have realistic portfolios.

agazso:
> @agazso you are right that the performance is not great (I don't experience this in my computer but it is surely on the upper bound performance wise). I gave claude some free reign with optimizing the new calculations and IMHO it improved things a lot. LMK what is your experience #874 .
> 
> I think we should generally optimize the calculations (because we unnecessarily recalculate things especially whole investments where nothing changed). But I think that should be a separate task which also adds performance metrics. For that it would be good to have realistic portfolios.

The improvements in #874 matter a lot, now turning the inflation on/off is almost instant. I suggest to merge it, then this one too and open a new issue about performance.

Issue #127, opened 2024-03-21
Stop calculation when the invested amount falls below 0


vojtechsimetka:
- show warning message above graph that portfolio dropped below 0 at certain date
- stop calculating all graph data once this happens - this is a problem that the user should see - somehow my portfolio reached negative amount, I need to fix it

david-gauquelin:
We need to discuss this further, adding the topic for our next meeting

david-gauquelin:
can happen when people create withdrawal that can't be done because not enough funds
can also happen when someone changes some deposits value

Need to display overlay or something over the value chart with tooltips
Need to display info when creating the transaction
Need to display also on the transaction charts (show how much can't be withdrawn)
Show also in the breakdown

agazso:
The calculation is stopping, but we want to show a warning if that happens. That issue is tracked as #948, #944 and #943

PR #842, opened 2025-08-26
fix: withdraw below zero

### LINKED ISSUE #100: Handle withdrawing more money than what the portfolio value is
- [ ] detect when the portfolio value dips below zero and show alert above the graph
- [ ] from the point when the portfolio is below zero, set the graph values should be all set to zero

### LINKED ISSUE #801: Withdrawals are still possible after investment value is zero
I generated a synthetic financial plan with Claude (https://github.com/snaha/kalkul/pull/800) and I found that the app still allows withdrawals after the investment value is zero. 

Image

### LINKED ISSUE #802: Fees can be positive in some cases
I generated a synthetic financial plan with Claude (https://github.com/snaha/kalkul/pull/800) and I found that fees can be positive in some cases. Maybe it is related to #801

Image

agazso:
This PR fixes situations where the investment value could go below zero. This is no longer possible and an error is reported when that would happen. This PR does not implement any UI change, but with the reported errors this is going to be possible once there is a UI design ready for this.

Fixes #801
Fixes #802 
Fixes #100 
Partially addresses #127


src/lib/@snaha/kalkul-maths/investment-calculations.test.ts
```diff
@@ -671,3 +680,198 @@ describe('#getCurrentPortfolioValue', () => {
 		expect(value).toBeGreaterThanOrEqual(0)
 	})
 })
+
+describe('Error scenarios - investment value below zero protection', () => {
```
vojtechsimetka: I am not too excited about this describe system but it seems AI has a tendency to write them: "error scenarios" or "edge cases"...

vojtechsimetka (approved):

src/lib/@snaha/kalkul-maths/investment-calculations.test.ts
agazso: I moved them under the `#getInvestmentValues` tests. The best is to tell Claude not to do that and then record that in the CLAUDE.md file.

PR #843, opened 2025-08-26
fix: cumulative interest

### LINKED ISSUE #840: Cumulative interest calculation is wrong
In the breakdown chart, the interest earned can decrease in time and even be negative. Most likely this is due to the withdrawals but clearly thje calculation is wrong.

Image

agazso:
This PR fixes the cumulative interest calculation. It also fixes another bug where the breakdown chart's tooltip also counted the hidden investments as part of the total but the chart was showing the correct number, so there was a discrepancy between the two.

Fixes #840


src/lib/components/graph-portfolio-breakdown.svelte
```diff
@@ -66,8 +66,19 @@
...
+		for (let i = 0; i <= index; i++) {
+			const valueChange = i === 0 ? values[i] : values[i] - values[i - 1]
+			const periodDeposits = deposits[i] || 0
+			const periodWithdrawals = withdrawals[i] || 0
+			const periodFees = fees[i] || 0
+
+			// Interest earned this period = value change - deposits - withdrawals - fees
+			const periodInterest = valueChange - periodDeposits - periodWithdrawals - periodFees
+			cumulativeInterest += periodInterest
+		}
```
vojtechsimetka: Nice, I wonder if we should not have this extracted somewhere with proper tests.

vojtechsimetka (approved):

src/lib/components/graph-portfolio-breakdown.svelte
agazso: I extracted the functions and added tests for it and it made me realize that the interest can go negative if the investment loses value. I opened an issue about this: #844

vojtechsimetka (approved):

PR #864, opened 2025-09-03
fix: settings back navigation

### LINKED ISSUE #860: Pressing the "back" button in Settings always brings back to clients list
I expect that it would bring me back to where I was

agazso:
Fixes #860

Also fixes layout issues on the "Change email" and "Change password" pages.


vojtechsimetka (approved):
awesome this was quite an annoyance when testing translations!

PR #869, opened 2025-09-03
fix: disable email obfuscation for any email we send

agazso:
This turns off the email obfuscation feature for all the emails that we send.


vojtechsimetka (approved):
I wonder id there is a way to disable this for specific emails only(support@kalkul.app) - because I am wary that there is chance we can leak emails we don't want to leak in future.

agazso:
> I wonder id there is a way to disable this for specific emails only([support@kalkul.app](mailto:support@kalkul.app)) - because I am wary that there is chance we can leak emails we don't want to leak in future.

There is no way to do this with Cloudflare based on the docs. We may add a check in the send-newsletter script.

PR #874, opened 2025-09-04
feat: optimize inflation adjustment transactions performance

vojtechsimetka:
This PR improves the performance by caching expensive operation and pre-calculating values.


agazso (approved):
Works well

agazso (commented):
This change seems to break the recalculation when changing the investment APY or just changing the transaction amount.

agazso:
Closed this one in favor of #910

PR #907, opened 2025-09-10
fix: update newsletter footer

agazso:


agazso:
This is included in #908

PR #917, opened 2025-09-12
fix: add busy state to create/edit buttons

### LINKED ISSUE #915: Create transaction button should be disabled after clicking
Currently creating a transaction may take quite long (>500ms) and the create button is not disabled after clicking it, so it can be clicked twice and then two transactions are created with the same parameters.

It would be better to disable the button (and maybe even add some spinner animation), so that it cannot be clicked more than one.

agazso:
Fixes #915


src/lib/utils.ts
```diff
@@ -71,3 +71,9 @@ export function parseLocalizedNumber(value: string, locale: string | null | unde
 
 	return parseFloat(normalized) || 0
 }
+
+export const asyncTimeout = (ms: number) => {
+	return new Promise((resolve) => {
+		setTimeout(resolve, ms)
+	})
+}
```
vojtechsimetka, line 79: This is just `sleep(ms: number)` function, right? I think we already have that.

src/lib/components/edit-client.svelte
```diff
@@ -35,7 +37,10 @@
 	let emailError: ZodFormattedError<z.infer<typeof emailFormSchema>> | undefined = $state()
 	let emailValid = $state(true)
 	let emailTouched = $state(false)
-	let createDisabled = $derived(name === '' || !birthDate || birthDate > date || !emailValid)
+	let createClicked = $state(false)
```
vojtechsimetka, line 40: I think we have a pattern for this - `isSaving`  or `isLoading`. WDYT?

vojtechsimetka (approved):

src/lib/utils.ts
agazso: In this project we don't have it.

src/lib/components/edit-client.svelte
agazso: We have `isLoading` but that is about displaying a spinner until certain stores are loaded. We can have an `isSaving` (which we don't at the moment). In that case I would recommend to create an issue and define the desired behavior.

PR #924, opened 2025-09-15
fix: improve password strength

### LINKED ISSUE #635: Display password requirements upfront on signup

### LINKED ISSUE #922: Improve password strenght requirements for account creation
Currently it's only 6 character long, let's make it better

agazso:
It would be better to not use Zod form validation at all, because when a required input is empty Zod returns `Required` as an error without localization. Also updated the form to match the design by the removing the checkbox for agreeing with terms and privacy policy.

Fixes #922
Fixes #635


agazso:
There is an issue when reloading the `/signup` page. The visible problem is that the text is originally rendered as czech, then it changes to english, except for the `agreeToTerms` text which stays czech.

Screenshot From 2025-09-15 16-44-45

agazso:
> There is an issue when reloading the `/signup` page. The visible problem is that the text is originally rendered as czech, then it changes to english, except for the `agreeToTerms` text which stays czech.

This was fixed with Claude adding a `{#key $locale}` directive around the problematic part.

src/lib/components/ui/input/input.svelte
```diff
@@ -13,7 +13,7 @@
 		dimension?: Dimension
 		layout?: Layout
 		unit?: string
-		error?: Snippet
+		error?: Snippet | string
```
vojtechsimetka, line 16: Nice. We have a lot of instances IMHO where we wrap string in a snippet just so that we can pass it in.

vojtechsimetka (approved):

PR #942, opened 2025-09-17
fix: enable analytics in instagram browser on iOS

### LINKED ISSUE #941: Analytics is disabled in Instagram in-app browser on iOS
See original issue #765 

Now there is a workaround deployed on production to avoid loading the analytics in Instagram in-app browser on iOS, but that means that we exclude people using that combination from our analytics data.

agazso:
This seems to fix #941 

The fix is that the analytics `script` is inserted after the content is loaded in the Instagram in-app browser on iOS.


vojtechsimetka (approved):

src/routes/+layout.svelte
```diff
@@ -72,3 +72,12 @@
 </svelte:head>
 
 {@render children()}
+
+<!-- Enable Umami Analytics on Instagram in-app browser on iOS -->
+{#if isProductionEnvironment() && isiOS() && isInstagramBrowser()}
```
vojtechsimetka, line 77: Wait it should actually disable it, right?

agazso: No, this is the fix for enabling analytics in the Instagram browser on iOS. For some reason it does not work if it's in the head (and it's disabled there), the fix is to insert the script after the content.

agazso: There were previous changes that I merged without review, so you might miss those:
#939 
#940

Issue #943, opened 2025-09-18
Design update for transactions/investment in sidebar (Inflation-adjusted + missing funds)


david-gauquelin:
This is a design update to display a badge and info about the inflation-adjusted feature and also indicate when there are missing funds.
Design details available here: https://www.figma.com/design/CH1QlTIGJk5xblFsNLi3Kn/Kalkul_05?node-id=9728-12466&t=K8WBOFNfbhKIXeJy-1

see also #127

PR #947, opened 2025-09-18
feat: consistent value details in transaction card component

vojtechsimetka:
Resolves partially #943


agazso (approved):
This looks good, but it's not for the "Add transaction" component footer, but it's about the transaction card (#943). Please update the title and the reference issue number.

PR #950, opened 2025-09-18
fix: newsletter subscription

### LINKED ISSUE #946: Subscribe to newsletter may fail
During signup or from the settings and since it is only reported on the console as error it is silently ignored.

agazso:
Fixes #946 

There is an improvement in error handling with `jsonError` to resemble more both the `json` function call and the `Error` constructor semantically. This allows richer data reporting when an error happens helping debugging issues.


src/routes/account/+page.svelte
```diff
@@ -141,7 +141,9 @@
 				await manageNewsletterSubscription(false)
 			}
 		} catch (e) {
-			console.error(`Error while ${newsletter_consent ? '' : 'un'}subscribing the newsletter`, e)
+			console.error(`Error while ${newsletter_consent ? '' : 'un'}subscribing the newsletter`, {
```
vojtechsimetka, line 144: Should we also display the error?

vojtechsimetka (approved):

src/routes/account/+page.svelte
agazso: Here `e` contains the error and also the cause, that is returned from the backend call.

Issue #963, opened 2025-09-25
Send email when someones tries to signup with an already registered email address


david-gauquelin:
we could send an email to that address saying that there's already an account registered with that address and include a link to sign up

Issue #957, opened 2025-09-23
Add show/hide feature for all password fields


david-gauquelin:
Since we decided to get rid of the "Confirm password" inputs we should add a show/hide button allowing people to see what they type

<img width="587" height="99" alt="Image" src="https://github.com/user-attachments/assets/805f6a7c-5bbe-4552-a005-6cb4a89216c3" />
<img width="595" height="113" alt="Image" src="https://github.com/user-attachments/assets/543e7805-b249-4bc4-b456-8f5001ea9977" />

Affected pages:

- Sign up
- Login
- Reset password
- Change email
- Change password

PR #966, opened 2025-09-26
fix: updated signup page

agazso:
This PR implements the updated signup page with the new design.

Figma reference: https://www.figma.com/design/j3GsCV6wdRAqiCeBf9mv9W/Kalkul_06?node-id=17-1621&t=OFPh1Jt8JPMpD7xW-4

Notes:
- The new behavior is to validate the form when the `Create account` button is pressed, in order to avoid unexpected layout changes
- The show/hide feature (#957) for the password field is implemented in a new `PasswordInput` component
- Instead of sending a new email it displays an error message when someone tries to signup with an existing email (#963), but now the error message is localized


package.json
```diff
@@ -9,6 +9,7 @@
 		"preview": "vite preview",
 		"test": "npm run test:integration && npm run test:unit",
 		"check": "svelte-kit sync && svelte-check --tsconfig ./tsconfig.json",
+		"check:all": "pnpm lint && pnpm check && pnpm knip && pnpm check-locales",
```
vojtechsimetka, line 12: Nice, we should also mention this in CLAUDE.md

vojtechsimetka (approved):

Issue #948, opened 2025-09-18
Show missing funds on charts


david-gauquelin:
<img width="1330" height="679" alt="Image" src="https://github.com/user-attachments/assets/79203444-e4a9-49f6-a1bc-a51dc0a9ff5c" />

These are ideas, not precise design — please check what's doable/easy in Chartjs based on this and let's iterate.

1. the label of the problematic investment should be displayed in red
2. we can add a line to indicate the problematic period, ideally it would be shown at the bottom of the graph, if not possible we could just color the zero axis line, but for the transactions chart this solution would look a bit weird (putting the red line in the middle of the chart)
3. we should add a badge a the beginning of the problem, again ideally I would position it above the line, but see what's doable and comment

david-gauquelin:
this should also include showing the error in the tooltip
I don't think we need a special tooltip for the error badge

<img width="561" height="280" alt="Image" src="https://github.com/user-attachments/assets/8c7ec1d9-0dc1-474a-a61a-58bc69c4d9d0" />

Issue #944, opened 2025-09-18
Design update for tooltips (inflation-adjusted and missing funds support)


david-gauquelin:
This is a design update that make sure that everything is consistent with the sidebar update #943 

Design details here: https://www.figma.com/design/CH1QlTIGJk5xblFsNLi3Kn/Kalkul_05?node-id=8665-48445&t=K8WBOFNfbhKIXeJy-1

david-gauquelin:
I also now added a small title to all tooltips to clarify the exact dates or period to help understand the data presented

PR #969, opened 2025-09-30
feat: show calculation errors on chart and in investment card

vojtechsimetka:
- [x] resolves https://github.com/snaha/kalkul/issues/948
- [x] resolves https://github.com/snaha/kalkul/issues/944
- [ ] partially implements https://github.com/snaha/kalkul/issues/943 (the warning on the inflation card) but does not show warning on the transaction card nor the missing amount. This will require a larger calculation rewrite which I'd prefer to do in a separate PR. Also please not that to show the warning on the inflation card adds computation overhead. The solution is to refactor the calculation to return problematic transactions for each investment and do this calculation outside of the chart components.

**Questions:**
1. If there is multiple investments in error, do we show just the first error icon on the chart? (current implementation)
2. Is there an error tooltip for the transaction chart? Right now it just stops showing the transactions for an investments that are in error state from the date this error occurred.


agazso (approved):
The code looks good, I found one potential issue. The warning sign is above a certain year and the red also starts from that year. However when hovering that year, there is no error displayed on the tooltip, only starting from the next year. Maybe this is by design, but I think it would be better to display the warning starting from the year when the funds are running out.

Screenshot From 2025-09-30 16-58-16

vojtechsimetka:
Good catch, I'll fix this

PR #1011, opened 2025-10-13
fix: retirement calculation

vojtechsimetka:


agazso (approved):
Looks good, tested it and it works well, though I did not review it line-by-line.

PR #1016, opened 2025-10-14
fix: mobile onboarding flow

### LINKED ISSUE #978: Portfolio page onboarding is not optimized for mobile
On mobile the helpbox is closed by default so only a disabled chart is shown. If the user finds the helpbox and opens it then the text inside is for the desktop version.

Image

Image

agazso:
Fixes #978

Also updated the helpboxes according to the latest design.


scripts/check-locales.ts
```diff
@@ -71,6 +117,7 @@ const EXCLUDE_PATTERNS = [
 	/^[a-zA-Z_$][a-zA-Z0-9_$.]*$/, // Simple variable names or property access
 	/^\w+\s*[<>]=?\s*\w+/, // Comparisons
 	/^=\s*\w+/, // Assignment operators or comparison starting with =
+	/^[A-Za-z0-9+/=]{20,}/, // Base64-like strings (20+ chars of base64 alphabet)
```
vojtechsimetka, line 120: This looks like it may catch any 20+char string, right?

vojtechsimetka (approved):

scripts/check-locales.ts
agazso: Yes, as long as it is a valid base64 encoded string (e.g. lowercase, uppercase, digits and `+`, `/` or `=`).

PR #1033, opened 2025-10-19
refactor: improved locale detections

vojtechsimetka:


eslint.config.mjs
```diff
@@ -23,6 +29,148 @@ export default typescriptEslint.config(
...
+						'beta',
+						'and',
+						'or',
+						'by',
+						'In',
+						'EN',
+						'CZ',
+						// UI labels
+						'Delete',
+						'Investments',
```
agazso, line 133: I think these should not be in this list

src/lib/components/ui/button.svelte
```diff
@@ -50,7 +49,7 @@
 	}: Props = $props()
 </script>
 
-<span class={`root ${className}`} class:disabled {style} class:flexGrow class:busy>
+<span class="root" class:disabled {style} class:flexGrow class:busy>
```
agazso, line 52: This change removes the ability to externally style buttons with classnames. There are examples where we are using it: in the custom-header and on the account page.

src/lib/components/ui/typography.svelte
```diff
@@ -60,7 +59,7 @@
 	class:line-through={lineThrough}
 	class:center
 	class:accent
-	class={`root ${font} ${variant} ${className}`}
+	class={`root ${font} ${variant}`}
```
agazso, line 62: This removes the ability to externally change the style or behavior with a classname and there are multiple examples for this, so this is a breaking change.

agazso (changes requested):
Please fix the  issues where I commented, the classname ones are breaking changes.

PR #1057, opened 2025-10-28
fix: chart hover during loading

### LINKED ISSUE #1002: Portfolio page breaks if tooltip hovered during loading
During the calculation of the chart data the charts are displaying a loading indicator. When hovered during this time a broken tooltip is shown and the calculations stops after that.

Most likely this can be fixed with disabling the tooltips during loading.

Image

agazso:
Fixes #1002

With the help of research from Claude, I found this issue: https://github.com/sveltejs/svelte/issues/14202 which is pretty much the same and suggested solution seems to be working for us as well.


vojtechsimetka (approved):
amazing, thank you!

PR #1064, opened 2025-10-29
feat: mcp server 

agazso:
This PR adds MCP server functionality so that LLM based AI can interact with the data. The operations exposed on the MCP interface are the same that are exposed to the user when interacting with the app.

Notes:
- The original version was built with Claude, which follows the protocol as described [here](https://modelcontextprotocol.io/docs/learn/server-concepts). It seems that OpenAI has a [different idea](https://platform.openai.com/docs/mcp) how MCP should work, so there is also support for `search` and `fetch` tools.
- Added an API token generation functionality with UI (with LLM agent design) that the MCP endpoint is using to authenticate the user. It can be used either as a bearer token or passed as query parameter (e.g. `/api/mcp?token={kalkul_XXX}`)
- There is a script for invoking the mcp commands locally and it is also available through `pnpm mcp:post` command for easier interfacing with CLI coding agents and tooling. It expects the `KALKUL_TOKEN` environment variable to be set to the API token value
- Renamed `CLAUDE.md` to `AGENTS.md` and linked `CLAUDE.md` and `GEMINI.md` to it

TODOs:
- [x] fix serviceAdapter session
- [ ] UI/design for the token management
- [ ] somehow notify the UI that the data was updated and needs to be reloaded
- [ ] #1070


vojtechsimetka (approved):

src/hooks.server.ts
```diff
@@ -23,8 +25,8 @@ export const handle: Handle = async ({ event, resolve }) => {
 		})
 	}
 
-	// For non-API routes
-	if (!event.url.pathname.startsWith(apiRoutes.ROOT)) {
+	// For non-API routes (check for /api/ with trailing slash to avoid matching /api-tokens)
+	if (!event.url.pathname.startsWith(apiRoutes.ROOT + '/')) {
```
vojtechsimetka: I'd prefer templating

src/hooks.server.ts
```diff
@@ -46,53 +48,132 @@ export const handle: Handle = async ({ event, resolve }) => {
...
-	if (
-		event.url.pathname !== apiRoutes.NEWSLETTER_SUBSCRIBE &&
-		!event.url.pathname.startsWith('/api/email')
-	) {
-		// For API routes, enforce Authorization header
+	// Skip auth for specific routes that don't require authentication
+	const isPublicRoute =
+		event.url.pathname === apiRoutes.NEWSLETTER_SUBSCRIBE ||
+		event.url.pathname.startsWith('/api/email') ||
+		(event.url.pathname === '/api/mcp' && event.request.method === 'GET')
```
vojtechsimetka: It would be good to have these as constants as well

src/lib/adapters/index.ts
```diff
@@ -1,4 +1,5 @@
 import type {
+	ApiToken, // New import
```
vojtechsimetka:
```suggestion
	ApiToken,
```

PR #1066, opened 2025-10-30
docs: goals implementation plan

vojtechsimetka:


docs/GOALS_IMPLEMENTATION_PLAN.md
```diff
@@ -0,0 +1,1186 @@
...
+
+**GitHub Issue**: [#1042](https://github.com/snaha-net/kalkul.app/issues/1042) - Introduce goal as investment
+
+**Goal**: Foundation + empty goals tab (feature flagged)
+
+**Changes:**
+
+- Add `goal_data` JSONB column to `investment` table
+- Create `goal_investment_link` table
+- Add database migration in `supabase/seed.sql`
```
agazso: For some reason LLMs want to add the database migration to seed.sql instead of creating a database migration file. I think it's worth adding this to CLAUDE.md

docs/GOALS_IMPLEMENTATION_PLAN.md
```diff
@@ -0,0 +1,1186 @@
...
+- No goal concept exists
+- Users manually create investments and transactions
+
+## Proposed Approach: Goals as Special Investments
+
+### Core Concept
+
+Store goals as investments with:
+
+- `type` column (existing): values like `'goal:retirement'`, `'goal:education'`, `'goal:house'`
```
agazso: Currently this is a plain text field that the user can edit. If we choose to repurpose this field then I would either remove it from the UI, or add a different field for the user editable content.

vojtechsimetka: It is editable but for investments, not for goals. There is no input where users could edit this.

agazso (approved):

docs/GOALS_IMPLEMENTATION_PLAN.md
agazso: Yes, this is what I meant, that it is editable for investments and we are adding a constraint to the investment table. This means that if a user writes `goal:house` in the type of their investment, there will be an error and won't be able to save the investment. I know this is unlikely but it's better to separate these things. I would either create a `goal` field for this or add an `investment_type` field for the user to edit.

vojtechsimetka: OK then we don't use that column and we can just put it directly in the JSON data, right?

agazso: > OK then we don't use that column and we can just put it directly in the JSON data, right?

That works, the goal types already have a `calculationType` field to discriminate on (I would simply call it `type` though)

PR #785, opened 2025-08-12
fix: database security + warnings

agazso:
This PR fixes the Supabase warning and addresses security and performance issues. It also removes the unused `feedback` table and related code.

The migration can be done as is, because it does not rely on new columns, therefore no new data needs to be added.

For now the authentication depends on the `advisor` column of the client and all the other tables are joined for checking. Later once there are teams the policy may involve another table (e.g. `team`) for authentication.

Also fixes an issue with anonymous view links that were an issue previously. It was caused by a combination of missing data from the read-only view and components assuming the stores are loaded with data even when using a share link anonymously.

Fixes #646
Fixes #787

PR #1072, opened 2025-10-31
feat: create and delete retirement type goals

### LINKED ISSUE #1042: Introduce goal as investment

### LINKED ISSUE #1043: Add retirement calculator

### LINKED ISSUE #1056: Add data colors to goals
Image

Image

My suggestion is to use the same palette than for investments, but in reverse order (starting from the end)
Carbon design system offers a default 'Categorical palette' with 14 colors.
https://carbondesignsystem.com/data-visualization/color-palettes/

Image

vojtechsimetka:
**What Users Can Do:**
  - Create retirement goals with customizable parameters (current savings, monthly deposits, retirement age, desired budget)
  - Preview retirement projections showing "What You Need" vs "What You Have" before saving
  - View saved goals in dedicated Goals tab with visual cards showing goal details and progress
  - See automatic transaction schedules (initial savings, recurring deposits, retirement withdrawals) on portfolio graph with goal-specific colors
  - Delete goals with confirmation modal

**Technical Implementation:**
  - Goals stored as investments with `goal_data` JSONB column (contains type: 'retirement' and calculation parameters)
  - Created `goal_investment_link` table for future investment-to-goal linking (Task 6)
  - Automatic transaction generation using `retirementGoalToTransactions()` with inflation adjustment
  - Separate portfolio simulations (`goalsSimulation` vs `investmentsSimulation`) for performance optimization
  - Feature-flagged rollout (`goals@kalkul.app`) with database migration (20251101091418_goals_schema.sql)
  - Retirement calculator logic optimized and moved to `src/lib/@snaha/kalkul-calculators/`

**Linked issues**
- Resolves #1042
- Resolves #1043
- Resolves #1056


src/lib/types.ts
```diff
@@ -66,3 +67,41 @@ export interface ApiToken {
 	user_id: string
 	token_hash: string // New field
 }
+
+// Goal types
+export type RetirementGoalData = {
+	type: 'retirement'
+	depositStart: string // ISO date
+	depositFrequency: 'month' | 'year'
```
agazso: I am not a fan of calling this frequency, because the dimension of frequency in physics is `1/time_unit` but the dimension of this is `time_unit`. So this is an inverse of what the name suggest.
In other places we called this periods or units.

src/lib/adapters/index.ts
```diff
@@ -52,6 +52,10 @@ export interface Adapter {
 	getInvestments: (portfolioId: number) => Promise<Investment[]>
 	getInvestment: (investmentId: number) => Promise<Investment | undefined>
 
+	// Goal methods (goals are investments with goal_data)
+	addGoal: (goal: Omit<Investment, MetaFields>) => Promise<number>
+	updateGoal: (goal: Partial<Investment> & Pick<Investment, 'id'>) => Promise<void>
+
```
agazso, line 58: If we want the AI to be able to work with this, we should also have a `getGoal(id)` and a `getGoals()` method. This can be done as a separate PR later.

vojtechsimetka: Let's do that as a separate PR but we can do it now so that this can be tested before we launch the AI feature

src/routes/(app)/client/[id]/portfolio/[portfolio_id]/goals/retirement/preview/+page@(app).svelte
```diff
@@ -0,0 +1,166 @@
...
+			budgetFrequency: calculatorData.calculationInput.budgetFrequency,
+			apy: calculatorData.calculationInput.apy,
+			inflation: calculatorData.calculationInput.inflation,
+		}
+
+		// Create goal in database as an investment with goal_data
+		const goalId = await adapter.addGoal({
+			portfolio_id: portfolioId,
+			name: $_('demo.investments.retirement'), // Default name - can be edited later
+			type: 'ETF', // Default type - can be edited later
```
Comment (agazso) [inline src/routes/(app)/client/[id]/portfolio/[portfolio_id]/goals/retirement/preview/+page@(app).svelte]: Why `ETF`?

supabase/migrations/20251101091418_goals_schema.sql
```diff
@@ -0,0 +1,75 @@
...
+GRANT TRUNCATE ON TABLE "public"."goal_investment_link" TO "authenticated";
+GRANT UPDATE ON TABLE "public"."goal_investment_link" TO "authenticated";
+
+GRANT DELETE ON TABLE "public"."goal_investment_link" TO "service_role";
+GRANT INSERT ON TABLE "public"."goal_investment_link" TO "service_role";
+GRANT REFERENCES ON TABLE "public"."goal_investment_link" TO "service_role";
+GRANT SELECT ON TABLE "public"."goal_investment_link" TO "service_role";
+GRANT TRIGGER ON TABLE "public"."goal_investment_link" TO "service_role";
+GRANT TRUNCATE ON TABLE "public"."goal_investment_link" TO "service_role";
+GRANT UPDATE ON TABLE "public"."goal_investment_link" TO "service_role";
```
agazso: These should be removed. `anon` should have no permission at all, there is no need for `service_role` permissions because that's the default and the RLS policy covers the authenticated case.

vojtechsimetka: I agree with you and I'll remove the "anon" permissions. But this is not that simple - the other tables also have `anon` access. I wanted to create an issue about this because we also need to keep in mind the View portfolio feature.

agazso: > I agree with you and I'll remove the "anon" permissions. But this is not that simple - the other tables also have `anon` access. I wanted to create an issue about this because we also need to keep in mind the View portfolio feature.

No, they don't have `anon` access (or if they have that's an error). You may be confused by the migration files that used to have those, but it was fixed after the RLS was introduced for all tables in #785. The view portfolio works by defining SQL functions that can be invoked with the `link_id` as argument and it joins the data appropriately. These functions are also defined in #785

agazso:
The mobile flow is inconsistent regarding goals. The button says `Show investments`, then it goes to the screen called `Investments` that shows both the goals & investments. There is a plus button on the top right which creates an investment, even when the view shows the goals.

Screenshot From 2025-11-06 15-14-37

Screenshot From 2025-11-06 15-14-45

agazso:
`No data` is shown at the place of the chart when there is no investments or goals are created.

Screenshot From 2025-11-06 15-16-51

agazso:
The helpbox is displayed inconsistently when there is no goal. It displays the same text as for the investment view.

Screenshot From 2025-11-06 15-19-39

agazso:
There is an unnecessay padding between the tab bar and the rest of the sidebar

Screenshot From 2025-11-06 15-33-39 (Copy) (Edit)

agazso (approved):
The spacing on the sidebar is still incorrect (should be 8px instead of 16px) but we can merge. Please open an issue about it. Another issue is to be opened is about the view link (besides the ones that I commented with screenshots).

PR #1116, opened 2025-11-19
feat: kid education calculator

agazso:
This PR implements the kid education calculator according to the latest design. It also implements the goal chooser screen.

This required quite a bit of a refactor, because the calculation part is 100% identical to the retirement, so instead of duplicating a lot of logic I extracted it to as `PeriodicWithdrawalCalculator`. Unfortunately this made the code more generic and therefore a bit more difficult to follow, but I felt it's a good tradeoff for avoiding duplication. There is also a generic component for the preview.

The goal definitions are also unified to extend a `PeriodicWithdrawalGoal` type. This changes the JSON definitions of the data types, including the one that is stored in the JSON field of the database, so this is a **breaking change** and existing data from the database will not work with this change. I think it's okay for now, because the feature is behind a feature switch, but we will have to keep an eye on this, because this is not enforced with tooling and does not require a database migration, but it still can break the app.


still-looking-for-name (changes requested):
1. The goal selection page extends to full width outside of the content of the rest of the page.
Screenshot 2025-11-30 at 12 22 42

2. The disabled goals have just part of the covering layout 
Screenshot 2025-11-30 at 12 22 28

3. Changing the Deposit amount does not change the Education budget (but this might be by design, it's just weird with these double sided arrows there)
Screenshot 2025-11-30 at 12 55 05

I also find it weird (but this is design decision) that on the form page we have a yearly budget and on the education preview page we have monthly budget.

agazso:
I fixed 1. and 2.

>     3. Changing the Deposit amount does not change the Education budget (but this might be by design, it's just weird with these double sided arrows there)

> I also find it weird (but this is design decision) that on the form page we have a yearly budget and on the education preview page we have monthly budget.

I agree, it is not completely clear what would be the best on these screen, most likely we would need to think more about it. For now we can also just remove the budget to simplify.

src/lib/components/goal-carousel.svelte
```diff
@@ -0,0 +1,135 @@
...
+		--vertical-gap="var(--padding)"
+		--vertical-align-items="center"
+		class="card"
+		onclick={() => {
+			if (!comingSoon) {
+				onSelect?.(name)
+			}
+		}}
+	>
+		<img src={image} alt="card" />
```
vojtechsimetka, line 31: This should be prefixed with a path. You can use `resolve` from `app/paths` (but maybe it needs update of sveltekit first see #1126)

https://svelte.dev/docs/kit/$app-paths#resolve

vojtechsimetka (approved):
Works great, thanks! The only issues are the images not being loaded on the preview pages.

agazso:
> Works great, thanks! The only issues are the images not being loaded on the preview pages.

Can you please clarify what exactly is the issue? I am not sure I understand.

Also is it a blocker for this calculator or can it be done as a separate PR?

vojtechsimetka:
Not a blocker, please merge

PR #1127, opened 2025-11-28
fix: newsletter subscribe issues

agazso:
This PR fixes issues related to resend.com API rate limit: https://resend.com/changelog/api-rate-limit


src/routes/api/newsletter/resend.ts
```diff
@@ -0,0 +1,12 @@
+import { RESEND_API_KEY } from '$env/static/private'
+import { asyncTimeout } from '$lib/utils'
+import { Resend } from 'resend'
+
+/**
+ * resend.com API has a rate limit of 2 requests per second
+ */
+export async function resendAPI_waitForRateLimit() {
+	await asyncTimeout(1_000)
+}
```
vojtechsimetka: Maybe we should also do a retries?

vojtechsimetka (approved):

src/routes/api/newsletter/resend.ts
agazso: I added a retry mechanism

